PR #16, opened 2019-07-18
Added Post processing scripts for smallfile

chawlanikhil24:
As discussed, to keep all the python files at 1 place i.e. snafu , I am adding the redis pubsub python scripts here. 
In addition to that, I am also adding a wrapper for smallfile to post process the data generated by smallfile.


smallfile-wrapper/publisher.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/python
+
+from __future__ import print_function
```
aakarshg: Just curious of the future function here. What's the reason behind the need for the future function here? trying to figure out which print() call in the script required the need for it.  or is the assumption here that the interpreter is python2 ?

smallfile-wrapper/publisher.py
```diff
@@ -0,0 +1,24 @@
...
+
+redis_host = sys.argv[1]
+number_of_pods = sys.argv[2]
+
+try:
+    r = redis.StrictRedis(host=redis_host, port=6379)
+    count = 0
+    while int(count) != int(number_of_pods):
+        count = r.execute_command("PUBSUB NUMSUB smallfile")[1]
+        print(count, file=sys.stdout)
```
aakarshg: just curious about the need the need to explicitly specify file as `sys.stdout`, because from my understanding the default file object to a print function us `sys.stdout` based on [documentation](https://docs.python.org/3/library/functions.html#print)

smallfile-wrapper/publisher.py
```diff
@@ -0,0 +1,24 @@
...
+        time.sleep(1)
+    print("All Pods are ready to run. Triggering the run ........\(^_^)/", file=sys.stdout)
```
aakarshg: same as above comment.

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,78 @@
...
+import sys
+import json
+
+
+# The purpose of this function is to consolidate all result jsons in to 1 list
+def _load_result_JSON(smallfile_path, operations):
+    results = []
+    for operation in operations:
+        file_path = smallfile_path + "_" + operation
+        file_path = "sample_smallfile_OP_from_a_Container"
```
aakarshg: looks like you're overwriting `file_path` variable here, probably not intended?

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,78 @@
...
+        with open(file_path) as json_file:
+            data = json.load(json_file)
```
aakarshg: Also just a suggestion but since you're already reading the json in here, maybe call the `_json_payload` function here and add the keys `workload`, `uuid` and `user` here itself? instead of loading the list of docs twice. Just an optimization suggestion here.

smallfile-wrapper/publisher.py
chawlanikhil24: It is because of the assumption that, by default, python2 interpreter will be invoked, when executing the script.

smallfile-wrapper/publisher.py
chawlanikhil24: I didn't know about that, for now , I am just replacing print() with sys,stdout.write() , keeping it irrespective of python version used.

smallfile-wrapper/publisher.py
chawlanikhil24: same as above

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: It was not at all intended, just added for testing purposes. Have to remove it.

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: Really a good idea, I'll right away implement the same. Thanks for pointing this optimisation.

smallfile-wrapper/publisher.py
bengland2: print() works fine in python3, just can't use it without parentheses

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,119 @@
...
+def main():
+
+    # retrieving elasticsearch server address and operating port from
+    # environement variables. The credentials are exported in the
+    # tasks of ripsaw/roles/smallfile-bench. These variables are exported
+    # based on condition, if the es_port and es_server is defined in
+    # smallfile CRDS of ripsaw. Use of "os.environ.get($env_var)" is to look
+    # for the supplied $env_var. If there is no such $env_var defined,
+    # then it will return None object.
+    server = os.environ.get("es")
```
jtaleric: we should  check to see if these exists prior to capturing them.

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,119 @@
...
+        # By reaching here, means the $uuid and $user has been successfully
+        # retrieved and injected in each piece of smallfile result. Now, data
+        # is all set to send to elasticsearch server. Checking for ES
+        # server and port address here. If available, _index_result() will s
+        # establish connection to ES server and send data, by iterating through
+        # each smallfile result in the "documents" list.
+        if (server is not None) and (port is not None):
+            if len(documents) > 0:
+                status = _index_result(server, port, documents)
+                if status:
```
jtaleric: _index_result only returns false if a failure happens, it never returns the status...

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,119 @@
...
+        with open(_path) as json_file:
+            data = json.load(json_file)
+            results.append(data)
+    return results
+
+
+# The purpose of this function is to establish connection with
+# elasticsearch server and index the consolidated results 1 by 1
+def _index_result(server, port, payload):
+    index = "smallfile-results"
```
jtaleric: ripsaw-smallfile-index

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,119 @@
...
+
+import elasticsearch
+import os
+import sys
+import json
+import traceback
+
+
+# The purpose of this function is to consolidate all result jsons in to 1 list
+def _load_result_JSON(smallfile_path, operations):
```
jtaleric: lowercase json :)

smallfile-wrapper/smallfile-wrapper.py
jtaleric: agree with @chawlanikhil24 ... Most of the SNAFU code has a method to assemble the json payload. This is just to read in the file (not sure why it needs to be its own function though).

smallfile-wrapper/smallfile-wrapper.py
```diff
@@ -0,0 +1,119 @@
...
+    port = os.environ.get("es_port")
+    # for each benchmark run, ripsaw generates a UUID and export it
+    # as a environement variable. Here, that uuid and a test_user is
+    # is trying to be fetched from env. Use of "os.environ.get($env_var)"
+    # is to look for the supplied $env_var. If there is no such $env_var
+    # defined, then it will return None object.
+    uuid = os.environ.get("uuid")
+    user = os.environ.get("test_user")
+
+    # Smallfile results path is based on the run, if it is executed on
```
jtaleric: lets use argparse so the user running this wrapper knows the inputs without having to read code.

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: I deliberately used os.environ.get("es"). Evenif, es doesn't exist, this will simply return None and i am checking for the None object later in the code.

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: ack

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: Thanks for pointing this out, adding a return statement , when its done.

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: sounds like a good idea. Will add it.

chawlanikhil24:
Code for redis is removed.

smallfile-wrapper/smallfile-wrapper.py
chawlanikhil24: Got a condition here, what sort of parameters we are looking for , that user can provide at the cli ? UUID, USER, ES_SERVER, PORT are captured through environment variables

jtaleric:
Title is misleading -- this is for smallfile, right?

jtaleric:
@chawlanikhil24 ?

chawlanikhil24:
@jtaleric is there any AI on me for this, else can we get it merged ?

chawlanikhil24:
rebased

acalhounRH:
Can we change the directory and file name to smallfile_wrapper, it would cut down on future changes to change all dashes to underscores.

chawlanikhil24:
@acalhounRH Can you please point me to the line of code, where you are suggesting the changes ?

bengland2:
@chawlanikhil24 I am working with Alex Calhoun to leverage run_snafu.py with smallfile and fs-drift, so that we don't have to copy-and-paste all the Elasticsearch-related code in every benchmark.    See run_snafu.py in development branch of snafu repo.   I'll try to take what you did here and incorporate it.  My branch (not yet a PR) is off the development branch, and is called smallfile-run-snafu.   I'll add one for fs-drift also.

aakarshg:
Closing this as changes have been merged through #41

PR #18, opened 2019-07-19
Adding reusable redis PUBSUB scripts

chawlanikhil24:
So, I converted the redis pubsub smallfile code to reusable, python PUBSUB code, which can executed with any available benchmarks in ripsaw.

### DIFF
--- a/redis_scripts/publisher.py
+++ b/redis_scripts/publisher.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python
+
+import redis
+import time
+import traceback
+import sys
+import argparse
+
+
+def run_publisher(redis_host, redis_port, benchmark, pod_count):
+    try:
+        r = redis.StrictRedis(host=redis_host, port=redis_port)
+        count = 0
+        while count != pod_count:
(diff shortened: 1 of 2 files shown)


chawlanikhil24:
@jtaleric @aakarshg @dustinblack

jtaleric:
This code looks good for the starting trigger.

redis_scripts/publisher.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def run_publisher(redis_host, redis_port, benchmark, pod_count):
+    try:
+        r = redis.StrictRedis(host=redis_host, port=redis_port)
+        count = 0
+        while count != pod_count:
+            redis_command = "PUBSUB NUMSUB {}".format(benchmark)
+            count = r.execute_command(redis_command)[1]
+            sys.stdout.write(str(count)+"\n")
```
jtaleric: sys.stdout.write ? Curious why you are using this?

chawlanikhil24:
changed sys.stdout.write() with print()

PR #20, opened 2019-07-24
fix indexing for conditions where keys are in global instead of job s…

dustinblack:
…ections of jobfile


aakarshg:
Like the try/catch but not sure its needed. the job dict can be updated with the keys in global dict thus the keys should be available.

dustinblack:
I needed to move forward, so I merged this. The try/except was the simplest
way I could find to untangle the functions while not affecting the existing
fio indexing. This should be a fine interim solution while waiting for the
new ripsaw PR to get merged.

PR #26, opened 2019-08-01
Added common logging and py_es_bulk

acalhounRH:
updated snafu directory to be pythonic, added common logging module, and integrated py_es_bulk.


aakarshg:
@acalhounRH needs a rebase

fio_wrapper/fio_wrapper.py
```diff
@@ -0,0 +1,163 @@
...
+    object_generator = process_data(args)
+    
+    for object in object_generator:
+        for action, index in object.emit_actions():
+            
+            es_valid_document = { "_index": index,
+                                  "_type": "result",
+                                  "_op_type": "create",
+                                  "_source": action,
+                                  "_id": "" }
```
aakarshg: why are we generating id here instead of having it generated directly on serverside?

utils/py_es_bulk.py
```diff
@@ -0,0 +1,197 @@
...
+# Use the random number generator provided by the host OS to calculate our
+# random backoff.
+_r = SystemRandom()
+# The maximum amount of time (in seconds) we'll wait before retrying an
+# operation.
+_MAX_SLEEP_TIME = 120
+# Always use "create" operations, as we also ensure each JSON document being
+# indexed has an "_id" field, so we can tell when we are indexing duplicate
+# data.
+_op_type = "create"
```
aakarshg: Not sure if this is absolutely needed, I don't see a way to index duplicate data. Theres a difference in indexing between pbench and snafu in that pbench has only one end point that does indexing, so you dont want to index the same tarball but with snafu the data is generated and then indexed, so we should remove this as its an unnecessary computation that we dont hace to do.

fio_wrapper/fio_wrapper.py
acalhounRH: If one does not generate their own id + use the create method elasticsearch will perform this action, thus causing an increase in computes to index on the ES server/cluster. using the create op_type and creating the id while indexing allows for faster indexing and eliminates the chance of duplication. regardless of likelihood of duplication, this approach safeguards against if there is ever a re-indexing of an archive.

utils/py_es_bulk.py
acalhounRH: Please see above response.

jtaleric:
I think I just caused the conflict w/ bakpack-wrapper -- sorry.

PR #29, opened 2019-08-02
Put each module from backpack/snafu to its own index

dry923:
Put each module from backpack/snafu to its own index. i.e. dmidecode-results, cpu_vulnerabilities-results, etc.

### DIFF
--- a/backpack-wrapper/backpack-wrapper.py
+++ b/backpack-wrapper/backpack-wrapper.py
@@ -30,7 +30,8 @@ def _index_result(server,port,payload_file):
     scribe_uuid = "NONE"
     for scribed in transcribe(payload_file,'stockpile'):
         try:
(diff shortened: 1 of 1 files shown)


backpack-wrapper/backpack-wrapper.py
```diff
@@ -30,7 +30,8 @@ def _index_result(server,port,payload_file):
     scribe_uuid = "NONE"
     for scribed in transcribe(payload_file,'stockpile'):
         try:
-            es.index(index=index, doc_type="result", body=scribed)
+            scribe_module = json.loads(scribed)['module']
+            es.index(index=scribe_module+"-results", doc_type="result", body=scribed)
```
jtaleric: module-metadata vs results?

dry923: Updated. Good call

PR #33, opened 2019-08-08
Added common logging and py_es_bulk

aakarshg:
Part 1 of #26 , which had too many conflicts


acalhounRH:
Need to change the directory name in order to be valid python syntax and import modules.

Change from:
snafu/fio-wrapper
to:
snafu/fio_wrapper

aakarshg:
> Need to change the directory name in order to be valid python syntax and import modules.
> 
> Change from:
> snafu/fio-wrapper
> to:
> snafu/fio_wrapper

Will do so in next iterations

PR #34, opened 2019-08-09
updating default index to send results to

aakarshg:


fio-wrapper/fio-wrapper.py
```diff
@@ -62,6 +62,9 @@ def main():
     args.index_results = False
     args.prefix = "ripsaw-fio-"
     es={}
+    args.cluster_name = "myk8scluster"
```
jtaleric: I would argue to set this to not myk8s -- simply due to your comment of having SNAFU be used in other areas.

aakarshg:
Made changes as requested.

PR #35, opened 2019-08-13
added conditional and log statemenet for yielding fio analyzer

acalhounRH:
This issue prevents unintended indexing of analyzed fio data.


.gitignore
```diff
@@ -0,0 +1,2 @@
+/.project
```
aakarshg: why is this necessary?

PR #40, opened 2019-09-06
add job options to results doc

dustinblack:


fio_wrapper/trigger_fio.py
```diff
@@ -52,6 +52,7 @@ def _document_payload(self, data, user, uuid, sample, list_hosts, end_time, fio_
               "fio-version": fio_version,
               "timestamp_end": int(end_time)*1000, #this is in ms
               #"nodeName": pod_details["hostname"],
+              "job_options": fio_jobs_dict[job],
```
aakarshg: 1. this wouldn't work, as no such variable job exists.
2. the variable result itself will contain the job specific options so this isn't necessary, the only thing result wouldn't contain would be the global options which if you look at L60 is taken care of

PR #41, opened 2019-09-06
Smallfile run snafu

bengland2:
This has been tested by running run_snafu.py directly with env. vars, but not within ripsaw.   That's next.   So you may want to hold off merging until it has run within a Ceph cluster and ripsaw.   But it appears to work correctly - data is now showing up in elastic search the way I expected it to, can anyone verify?  I also plan to add response time data in the future but let's just get the basics in place.  I pushed changes into smallfile to make its JSON output elasticsearch-friendly (no lists, everything is dictionary field, date field).

### DIFF
--- a/smallfile_wrapper/trigger_smallfile.py
+++ b/smallfile_wrapper/trigger_smallfile.py
@@ -92,7 +92,6 @@ class _trigger_smallfile:
             with open(rsptime_file) as rf:
                 lines = [ l.strip() for l in rf.readlines() ]
                 start_grabbing = False
-                rsp_percentiles = {}
                 for l in lines:
                     if l.startswith('time-since-start'):
                         start_grabbing = True
@@ -103,6 +102,7 @@ class _trigger_smallfile:
                         interval = {}
                         rsptime_date = start_time + int(flds[0])
                         rsptime_date_str = time.strftime('%Y-%m-%dT%H:%M:%S.000Z', time.gmtime(rsptime_date))
+                        interval['date'] = rsptime_date_str
                         interval['samples'] = int(flds[2])
                         interval['min'] = float(flds[3])
                         interval['max'] = float(flds[4])
@@ -111,5 +111,4 @@ class _trigger_smallfile:
                         interval['90%'] = float(flds[8])
                         interval['95%'] = float(flds[9])
                         interval['99%'] = float(flds[10])
-                        rsp_percentiles[rsptime_date_str] = interval
-                yield rsp_percentiles, '-rsptimes'
+                        yield interval, '-rsptimes'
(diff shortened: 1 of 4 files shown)


smallfile_wrapper/trigger_smallfile.py
```diff
@@ -0,0 +1,114 @@
...
+                process = subprocess.check_call(cmd, stderr=subprocess.STDOUT)
+            except subprocess.CalledProcessError as e:
+                self.logger.exception(e)
+                raise SmallfileWrapperException(
+                    'smallfile_cli.py non-zero process return code %d' % e.returncode)
+            self.logger.info("completed sample {} for operation {} , results in {}".format(
+                        self.sample, operation, json_output_file))
+            with open(json_output_file) as f:
+                data = json.load(f)
+                yield data, '-results'
```
aakarshg: Do you've an example json document @bengland2 so that we can see what's the structure of the document.

smallfile_wrapper/trigger_smallfile.py
```diff
@@ -0,0 +1,114 @@
...
+                        rsptime_date_str = time.strftime('%Y-%m-%dT%H:%M:%S.000Z', time.gmtime(rsptime_date))
+                        interval['date'] = rsptime_date_str
+                        interval['samples'] = int(flds[2])
+                        interval['min'] = float(flds[3])
+                        interval['max'] = float(flds[4])
+                        interval['mean'] = float(flds[5])
+                        interval['50%'] = float(flds[7])
+                        interval['90%'] = float(flds[8])
+                        interval['95%'] = float(flds[9])
+                        interval['99%'] = float(flds[10])
```
aakarshg: we'll want to add some additional identifiers such as the uuid clustername testuser and so on, ptal at the other benchmarks say uperf or ycsb.

smallfile_wrapper/trigger_smallfile.py
bengland2: [here you go](http://perf1.perf.lab.eng.bos.redhat.com/pub/bengland/tmp/example-smallfile.json)

smallfile_wrapper/trigger_smallfile.py
bengland2: oh, so all of that goes in every document?   Is that inefficient?  Will look at fio and see how it did it.  That explains why I can't see the data in Kibana yet perhaps.

aakarshg: yes that's right, and it's not inefficient. You can think of them as primary/secondary keys that are necessary to identify each entry.

aakarshg: not inefficient, as they're essentially the primary/secondary keys that help us identify each record

PR #43, opened 2019-09-11
Fs drift run snafu

bengland2:
fs-drift is integrated with run_snafu.py in the same way as smallfile, basically.  I modified fs-drift to incorporate date field into JSON output so nothing had to be done there, but I added date field to response time time-series docs.


acalhounRH:
Check this out, good to go.

aakarshg:
Updated the PR to point at :master now :)

PR #45, opened 2019-09-13
Adding dockerfiles from magazine into snafu

aakarshg:
This is due to the following reasons:
1. magazine can't exist on its own, as the user would need to understand how to trigger it
2. reduces overhead of having to update image in magazine when snafu is updated


bengland2:
LGTM.

PR #49, opened 2019-09-17
added function to add KiB io size to all analyzed documents.

acalhounRH:
This is needed to better graph data in Grafana

### DIFF
--- a/fio_wrapper/fio_analyzer.py
+++ b/fio_wrapper/fio_analyzer.py
@@ -30,6 +30,66 @@ class Fio_Analyzer:
             fio_result["starttime"] = starttime
             self.fio_processed_results_list.append(fio_result)
 
+    def standardize_units(self, io_size):
+        """
+            Will return value converted to Kib for known io sizes without unit size suffix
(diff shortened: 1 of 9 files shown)


aakarshg:
added @dustinblack to review, since you've knowledge of units in fio.

acalhounRH:
Will update to handle MB as well, what would be the handling for kb_base?

acalhounRH:
Updated to handle edge case of switching kb_base value from 1000 to/from 1024. and handling of IEC and non-IEC values.

aakarshg:
@acalhounRH this needs a  rebase.

dry923:
@acalhounRH can you rebase this or would you rather close and open a new one when you get back to it?

jtaleric:
@dry923 @aakarshg i suggest if we don't hear anything today (Feb 18th) .We close, Alex might be busy with other things ATM.

aakarshg:
> @dry923 @aakarshg i suggest if we don't hear anything today (Feb 18th) .We close, Alex might be busy with other things ATM.

agreed closing this out for now.

PR #51, opened 2019-09-17
Fs drift dockerfile

bengland2:
This has been tested in AWS cluster, previously I wasn't testing run_snafu.py from inside a container built with the Dockerfile.     Note that this depends on pip instead of RPM now and python scripts that are run by trigger are assumed to be in PATH (Dockerfile does this).


bengland2:
I modified to use COPY instead of snafu clone.    Again, this works in AWS Ceph cluster now, except for sending ES data back to marquez.

PR #52, opened 2019-09-18
Added start time to fio-results document

acalhounRH:


fio_wrapper/trigger_fio.py
```diff
@@ -55,12 +55,15 @@ def _document_payload(self, data, user, uuid, sample, list_hosts, end_time, fio_
             }
             if 'global' in fio_jobs_dict.keys():
                 document['global_options'] = fio_jobs_dict['global']
-            processed.append(document)
+            
             if result['jobname'] != 'All clients':
                 start_time= (int(end_time) * 1000) - result['job_runtime']
                 fio_starttime[result['hostname']] = start_time
+                formatted_starttime = datetime.fromtimestamp(start_time / 1000)
+                document['start_time'] = formatted_starttime.strftime('%Y-%m-%dT%H:%M:%S.%fZ')
```
aakarshg: 1. We should index timestamps as they are returned ( /1000 is fine as they're still in epoch format ) 
2. We should be using the earliest_starttime as the value for start time to be indexed.

aakarshg:
closing this in favor of #63

PR #58, opened 2019-09-30
hammerdb specifics

mkarg75:
Changed the variables that are actually used. This still will need some trimming when it comes to indexing, but for now it should be good.


hammerdb-wrapper/hammerdb-wrapper.py
```diff
@@ -0,0 +1,167 @@
...
+    return stdout.strip(), process.returncode
+
+def _fake_run():
+    with open("hammerdb.log", "r") as input:
+        stdout = input.read()
+    return stdout,0
+
+def _parse_stdout(stdout):
+    data = []
+    iteration = 0
```
jtaleric: Where is iteration being used in this function?

jtaleric: @mkarg75

mkarg75: It's not used at all - hammerdb itself has a parameter called iteration which I originally wanted to use but turned out to be not what its name suggests. I'm in touch with the hammerdb author and maintainer to clarify a few gaps I have in understanding some of the available settings. Will clean out whatever is no longer needed in the code, once that's done.

PR #59, opened 2019-10-01
Add Latency numbers to Uperf results

mohit-sheth:
This commit will add latency(microseconds) in addition to the already displayed throughput and ops/second result
@jtaleric

### DIFF
--- a/uperf-wrapper/uperf-wrapper.py
+++ b/uperf-wrapper/uperf-wrapper.py
@@ -1,5 +1,4 @@
-#!/usr/bin/env python
-#   Licensed under the Apache License, Version 2.0 (the "License");
+Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


uperf-wrapper/uperf-wrapper.py
```diff
@@ -133,6 +144,19 @@ def _summarize_data(data):
                              np.median(op_result),
                              np.average(op_result),
                              np.percentile(op_result, 95)))
+
+    print("")
+    print("UPerf Latecny results (microseconds):")
```
aakarshg: we're multiplying by 1000 on L40, wanted to confirm if the timestamp being returned was in milliseconds. Also a minor nit with the s/Latecny/Latency :)

mohit-sheth: yes, timestamp was in milliseconds

PR #64, opened 2019-10-08
Additional changes to the hammerdb addition

mkarg75:
Had to fix a few bits and pieces, should only affect hammerdb


aakarshg:
@mkarg75 can you please squash the commits into one and rebase against the master so that its easier to see the differences that this pr actually introduces?

hammerdb-wrapper/hammerdb-wrapper.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python
```
aakarshg: The file that needs to be updated is https://github.com/cloud-bulldozer/snafu/blob/master/hammerdb/hammerd_wrapper.py, also i think you'd mentioned a need to change the dockerfile as well. Lmk once its ready :)

hammerdb/Dockerfile
```diff
@@ -4,24 +4,28 @@ FROM centos/tools
...
 RUN yum remove -y unixODBC-utf16 unixODBC-utf16-devel && yum -y clean all
-RUN yum clean all -y
+RUN yum clean all -y 
 
-# clone the snafu repo
-#RUN git clone https://github.com/mkarg75/snafu.git
-# Download and install the hammer suite
+# clone the snafu repo 
+#RUN cd /opt/; git clone https://github.com/mkarg75/snafu.git
+RUN cd /opt/; git clone https://github.com/cloud-bulldozer/snafu.git
```
aakarshg: since this dockerfile is inside snafu, we don't have to clone it but rather just move(copy) it and this can be done as follows:
```
RUN mkdir -p /opt/snafu/
COPY . /opt/snafu/
```

hammerdb/Dockerfile
```diff
@@ -4,24 +4,28 @@ FROM centos/tools
...
 
+# Download and install the hammer suite
 RUN wget 'https://downloads.sourceforge.net/project/hammerdb/HammerDB/HammerDB-3.2/HammerDB-3.2-Linux-x86-64-Install?r=https%3A%2F%2Fsourceforge.net%2Fprojects%2Fhammerdb%2Ffiles%2FHammerDB%2FHammerDB-3.2%2FHammerDB-3.2-Linux-x86-64-Install%2Fdownload&ts=1564587940&use_mirror=autoselect' -O hammer_installer
 RUN mkdir /hammer
 RUN chmod 755 hammer_installer
 RUN ./hammer_installer --prefix /hammer --mode silent
 
-COPY hammerdb/uid_entrypoint /usr/local/bin/
+COPY uid_entrypoint /usr/local/bin/
```
aakarshg: this will need to be `hammerdb/uid_entrypoint` as the build context would be root fo repository, so that we can copy the repo as in the above comment.

hammerdb/hammerd_wrapper.py
```diff
@@ -147,16 +149,15 @@ def main():
     stdout = _run_hammerdb()
     #stdout = _fake_run()
     if stdout[1] == 1:
-        print "hammerdbcli failed to execute, trying one more time.."
+        print ("hammerdbcli failed to execute, trying one more time..")
         stdout = _fake_run()
```
aakarshg: this needs to change as well.

hammerdb/hammerd_wrapper.py
```diff
@@ -106,7 +108,7 @@ def main():
     args = parser.parse_args()
 
     #server = "bullwinkle-elk.rdu.openstack.engineering.redhat.com"
-    server = ""
+    server = "marquez.perf.lab.eng.rdu2.redhat.com"
```
aakarshg: we should be able to read this from environment variables.

hammerdb/hammerd_wrapper.py
```diff
@@ -24,34 +24,38 @@ def _parse_stdout(stdout):
     data = []
     iteration = 0
     for line in stdout.splitlines():
+        print (line)
         if "TEST RESULT" in line:
             worker = (line.split(":"))[0]
             tpm = (line.split(" "))[6]
             nopm = (line.split(" "))[-2]
-            entry = [ worker, tpm, nopm]
+            timestamp = datetime.now()
```
aakarshg: so I think we should be getting the timestamp just before we execute the hammerdb command vs while parsing the output

PR #65, opened 2019-10-08
Fixed es indexing for hosts with auth

aakarshg:
Changes needed to use es client where the server has basic auth, the current way of initiating through host and port, won't work.
Also, removed using type, and fixed some more es related misc issues.

### DIFF
--- a/hammerdb/hammerd_wrapper.py
+++ b/hammerdb/hammerd_wrapper.py
@@ -89,12 +89,11 @@ def _summarize_data(data):
               NOPM: {}""".format(entry['nopm']))
         print("+{}+".format("-"*(115)))
 
-def _index_result(server, port, payload):
-    index = "hammerdb-results"
-    es = elasticsearch.Elasticsearch([
-        {'host': server, 'port': port}], send_get_body_as='POST')
+def _index_result(index,server,port,payload):
+    _es_connection_string = str(host) + ':' + str(port)
+    es = elasticsearch.Elasticsearch([_es_connection_string],send_get_body_as='POST')
     for result in payload:
-        es.index(index=index, doc_type="result", body=result)
+        es.index(index=index, body=result)
 
 def main():
     parser = argparse.ArgumentParser(description="HammerDB Wrapper script")
@@ -145,7 +144,7 @@ def main():
         samples = os.environ["samples"]
 
 
-    stdout = _run_hammerdb() 
+    stdout = _run_hammerdb()
     #stdout = _fake_run()
     if stdout[1] == 1:
         print "hammerdbcli failed to execute, trying one more time.."
@@ -158,7 +157,7 @@ def main():
     if server != "" :
         if len(documents) > 0 :
             print "Indexing data"
-            _index_result(server,port,documents)
+            _index_result("ripsaw-hammerdb-results",server,port,documents)
     if len(documents) > 0 :
         _summarize_data(documents)
 
--- a/run_snafu.py
+++ b/run_snafu.py
@@ -38,7 +38,7 @@ urllib3_log.setLevel(logging.CRITICAL)
 setup_loggers("snafu", logging.DEBUG)
 
 def main():
-    
+
     #collect arguments
     parser = argparse.ArgumentParser(description="run script")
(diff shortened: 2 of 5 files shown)


aakarshg:
@bengland2 please review

run_snafu.py
```diff
@@ -84,27 +84,26 @@ def main():
...
     tdelta = end_t - start_t
     logger.info("Duration of execution - %s" % tdelta)
 
 
 
 def process_generator(index_args, parser):
-    
+
     benchmark_wrapper_object_generator = generate_wrapper_object(index_args, parser)
-
```
bengland2: I think these whitespace changes are caused by Alex using Eclipse as his editor.   Did you make these changes yourself?

hammerdb/hammerd_wrapper.py
```diff
@@ -89,12 +89,11 @@ def _summarize_data(data):
               NOPM: {}""".format(entry['nopm']))
         print("+{}+".format("-"*(115)))
 
-def _index_result(server, port, payload):
-    index = "hammerdb-results"
-    es = elasticsearch.Elasticsearch([
-        {'host': server, 'port': port}], send_get_body_as='POST')
+def _index_result(index,server,port,payload):
+    _es_connection_string = str(host) + ':' + str(port)
+    es = elasticsearch.Elasticsearch([_es_connection_string],send_get_body_as='POST')
```
bengland2: If we're going to convert to run_snafu.py then this doesn't matter, but you didn't catch exception here, but you do in run_snafu.

run_snafu.py
```diff
@@ -48,30 +48,30 @@ def main():
...
-    
-        es = elasticsearch.Elasticsearch([
-        {'host': es['server'],'port': es['port'] }],send_get_body_as='POST')
-    
-        if not es.ping():
-            logger.warn("Elasticsearch connection failed or passed incorrectly, turning off indexing")
+        try:
+            _es_connection_string = str(host) + ':' + str(port)
+            es = elasticsearch.Elasticsearch([_es_connection_string],send_get_body_as='POST')
+            logger.info("Connected to the elasticsearch cluster with info as follows:" + str(es.info()))
```
bengland2: maybe I'm slow, but the title of the commit mentioned "with auth", I don't see that part.

run_snafu.py
```diff
@@ -48,30 +48,30 @@ def main():
...
+        except Exception as e:
```
bengland2: I don't think the exception will be thrown here.  It will be thrown when you try to post results to an index.   I commented out the es.ping() call in run_snafu and that's what happened.   That's why Alex was calling es.ping(), to try to find out if there really was a connection or not, right?

run_snafu.py
aakarshg: Nope i didn't make the changes, so yeah looks like it could be the reason maybe?

hammerdb/hammerd_wrapper.py
aakarshg: This is a temporary fix in place, until we move to run_snafu. hence didn't add the try catch block, also as it happens creating an object will succeed even if given insane credentials so the try catch needs a command like es.info() to check, which again changes depends on the role. given that for our use case, we'll be moving towards run_snafu it didn't make a lot of sense to put all the checks in place here.

run_snafu.py
aakarshg: Updated the message my bad, missed it. The current manner of creating a elasticsearch object where the server has basic auth doesn't work, we'd haev to either use additional parameters like username and password or have it in connection url. connection url made most sense as it'd allow for usecases with reverse proxy too.

run_snafu.py
aakarshg: the exception will be thrown here when es.info() doesn't work, and in our case i gave the user snafu monitoring permissions too, so it'll help us catch the issue if something changes. making es.ping() work requires additional permissions to the snafu user, which would complicate things, vs for es.info() which only requires monitoring permission.

run_snafu.py
aakarshg: Yep, i alternate between vim and atom so looks like it

PR #66, opened 2019-10-08
do not report rsptime stats on interval with 0 iops

bengland2:
Without this change, the smallfile wrapper gets an exception converting empty string to floating point number if zero IOPS during an interval.   The record in rsptimes CSV output file is basically empty after the samples column, which contains a zero.


bengland2:
@aakarsh, I hit this problem while testing in AWS, need to have this PR merged.   Can you make it "OK to test" so we can run it through CI?  Should have brought this up at meetings but we were focused on AWS ES server.

PR #67, opened 2019-10-08
generate new index showing fs-drift throughput rates as a function of time

bengland2:
This allows fs-drift wrapper to convert each periodically reported counter into a rate so that we can graph them and see how different event rates changed over time during the test.    This is important for understanding results of this benchmark, which can run for hours or even days.   The user can control reporting rate to prevent it from overloading ES with data.   Perhaps we need to automatically increase reporting interval for long tests to something like 1% of test duration by default?


aakarshg:
hi @bengland2 do you've example documents for this?

bengland2:
good point because I need to apply the lessons from smallfile to this set of documents.   Will review again.   Do not merge yet.

aakarshg:
@bengland2 this needs a rebase so that we cna get the ci testing.

bengland2:
rebased, tested today with just fs_drift_wrapper/ci_test.sh, it passed.  Please label this one OK TO TEST

fs_drift_wrapper/trigger_fs_drift.py
```diff
@@ -130,3 +133,70 @@ def emit_actions(self):
...
+                        rate_obj['date'] = timestamp_str
+
+                        # add other info needed to display data in elastic search
+
+                        rate_obj['thread'] = thread_id
+                        rate_obj['cluster_name'] = self.cluster_name
+                        rate_obj['user'] = self.user
+                        rate_obj['uuid'] = self.uuid
+                        rate_obj['sample'] = self.sample
+                        yield rate_obj, '-rates-over-time'
```
aakarshg: this should `rates-over-time`

aakarshg:
has a conflict in fs_drift_wrapper/trigger_fs_drift.py can you please fix it @bengland2 ? CI won't be able to run if there's a merge conflict

bengland2:
I rebased, successfully this time.

bengland2:
can we mark this OK to test so that I can try it in CI?  I think it's ready to go, shouldn't impact any other benchmark.

fs_drift_wrapper/trigger_fs_drift.py
aakarshg: @bengland2 after the change the starting hyphen needs to be removed

bengland2:
ouch, should have rebased it better.   I know what you mean, will get on it as soon as I finish review of Raul's smallfile PR am trying to run that now.

bengland2:
I changed fs-drift so [default is to always generate rates-over-time json files](https://github.com/parallel-fs-utils/fs-drift/commit/a7b5ef56cf30901c162ff41f960f87427ff67a8a), so fs-drift doesn't have to pass --report-interval parameter (unless user wants to).

bengland2:
oops, some problems with this one, do not merge yet.

fs_drift_wrapper/trigger_fs_drift.py
bengland2: fixed

bengland2:
now the 3rd index works and should be ready to merge, I ran the CI test from my laptop successfully (using PR 104!)  and saw the data output with Kibana.

bengland2:
and [here is a grafana dashboard](http://marquez.perf.lab.eng.rdu2.redhat.com:3000/d/sZqsx3sWz/fs-drift-internal?fullscreen&edit&panelId=2&from=1580408904447&to=1580409643455) for it

bengland2:
please merge if no issues, this change only affects fs-drift

PR #71, opened 2019-10-12
added more details about time/date field

aakarshg:


README.md
```diff
@@ -78,9 +78,13 @@ server that is viewable with Kibana and Grafana!
 
 Look at some of the other benchmarks for examples of how this works.
 
+Note: Please tag @aakarshg for review when adding a new benchmark wrapper especially ones that index data into elasticsearch and mention:
```
bengland2: ok for now but long-term we can't have you responsible for every single benchmark getting a mapping in every single ES server.   Is there a way you can train the rest of us to help do this?

README.md
```diff
@@ -78,9 +78,13 @@ server that is viewable with Kibana and Grafana!
...
+* which field(key) in the document will be of the type date(time)
+* highly recommend that the key of the type date(time) use epoch format (helps a lot when dealing with tool run in multiple timezones like snafu)
```
bengland2: Which benchmark does this today?   If none, why don't we start by converting some benchmark to use this if we haven't already, and verify that it works, before we make this recommendation?  I volunteer to use fs-drift for this, because it's not hard to change and won't impact at lot of people so far.   If this one works then I can help with the other ones.

README.md
jtaleric: Agreed. This isn't maintainable.

README.md
jtaleric: This is how Scribe is indexing today. I just shared with @dry923 how our templates need to be on the ES side.

bengland2: Where is the Scribe code that does it?  Maybe we could make a short example illustrating how to do it from this code.

README.md
bengland2: Isn't there a way in Kibana to define this?  Or does that only apply to using Kibana?

README.md
aakarshg: upto the user but it can be done as follows: `int(time.time())`

aakarshg:
Agreed this is not maintainable, will work towards automation in subsequent sprints, can we merge this so at least some documentation is there, and we cna improve upon it later.

bengland2:
In hindsight I would have implemented all my snafu wrappers using int(time.time()) much easier.  I may go back and retrofit them if time permits, but this would break the kibana index patterns I've established, so I'd have to fix them too etc. @acalhounRH ^   Wish we could have more OO code pattern where we inherit from a base class that does the wrapper-independent stuff in one place.

aakarshg:
closing this, and will implement a process to create mapping over manual creation

PR #74, opened 2019-10-17
update fio wrapper to support greater than 1 numjobs

acalhounRH:
This update is to allow users to specify greater than 1 numjobs. removed greater than 1 check, and added a loop for processing log file for N number of numjobs. 

@aakarshg 
@dustinblack 
@jtaleric 
@bengland2

### DIFF
--- a/fio_wrapper/trigger_fio.py
+++ b/fio_wrapper/trigger_fio.py
@@ -79,44 +79,54 @@ class _trigger_fio:
             del _current_log_files['clat']
         if 'disable_bw' in job_options or 'disable_bw_measurement' in job_options:
             del _current_log_files['bw']
+        #find the number of jobs either in the job options or global options
+        if 'numjobs' in job_options:
+            numjob_list = job_options['numjobs']
+        else:
+            numjob_list = fio_jobs_dict['global']['numjobs']
+            
         for log in _current_log_files.keys():
             for host in list_hosts:
-                log_file_prefix_string = 'write_' + str(log) + '_log'
-                if log in ['clat','slat']:
-                    log_file_prefix_string = 'write_lat_log'
-                try:
-                    log_file_name = str(job_options[log_file_prefix_string]) + '_' + str(log) + '.1.log.' + str(host)
-                except KeyError:
+                for numjob in range(int(numjob_list)):
+                    numjob = numjob + 1
+                    log_file_prefix_string = 'write_' + str(log) + '_log'
+                    if log in ['clat','slat']:
+                        log_file_prefix_string = 'write_lat_log'
                     try:
-                        log_file_name = str(fio_jobs_dict['global'][log_file_prefix_string]) + '_' + str(log) + '.1.log.' + str(host)
-                    except:
-                        logger.info("Error setting log_file_name")
-                with open(directory+'/'+str(log_file_name), 'r') as log_file:
-                    for log_line in log_file:
-                        log_line_values = str(log_line).split(", ")
-                        if len(log_line_values) == 4:
-                            timestamp_ms = int(fio_starttime[host]) + int(log_line_values[0])
-                            newtime = datetime.fromtimestamp(timestamp_ms / 1000.0)
-                            log_dict = {
-                              "uuid": uuid,
-                              "user": user,
-                              "host": host,
-                              "cluster_name": self.cluster_name,
-                              "fio-version": fio_version,
-                              "job_options": job_options,
-                              "job_name": str(job),
-                              "sample": int(sample),
-                              "log_name": str(log),
-                              "timestamp": timestamp_ms, #this is in ms
-                              "date": newtime.strftime('%Y-%m-%dT%H:%M:%S.%fZ'),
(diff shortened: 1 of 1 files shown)


fio_wrapper/trigger_fio.py
```diff
@@ -79,44 +79,51 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
...
+            
         for log in _current_log_files.keys():
             for host in list_hosts:
-                log_file_prefix_string = 'write_' + str(log) + '_log'
-                if log in ['clat','slat']:
-                    log_file_prefix_string = 'write_lat_log'
-                try:
-                    log_file_name = str(job_options[log_file_prefix_string]) + '_' + str(log) + '.1.log.' + str(host)
-                except KeyError:
+                for numjob in numjob_list:
```
aakarshg: numjob_list would contain an integer like `4` or such, but python being python would assume the type is a string, so convert it to '4' and this for loop would only iterate over the value 4, but not from [1-4] which is what we want.

fio_wrapper/trigger_fio.py
```diff
@@ -79,44 +79,51 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
...
+                        for log_line in log_file:
+                            log_line_values = str(log_line).split(", ")
+                            if len(log_line_values) == 4:
+                                timestamp_ms = int(fio_starttime[host]) + int(log_line_values[0])
+                                newtime = datetime.fromtimestamp(timestamp_ms / 1000.0)
+                                log_dict = {
+                                  "uuid": uuid,
+                                  "user": user,
+                                  "host": host,
+                                  "cluster_name": self.cluster_name,
```
aakarshg: we'd also want to index which job does this particular metric refer to.

fio_wrapper/trigger_fio.py
```diff
@@ -79,44 +79,51 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
...
+                    except KeyError:
+                        try:
+                            log_file_name = str(fio_jobs_dict['global'][log_file_prefix_string]) + '_' + str(log) + '.' + numjob + '.log.' + str(host)
+                        except:
+                            logger.info("Error setting log_file_name")
+                    with open(directory+'/'+str(log_file_name), 'r') as log_file:
+                        for log_line in log_file:
+                            log_line_values = str(log_line).split(", ")
+                            if len(log_line_values) == 4:
+                                timestamp_ms = int(fio_starttime[host]) + int(log_line_values[0])
```
aakarshg: not sure if this is right always, this means we're assuming that all the jobs start at the exact same moment. have to test it to make sure its right, i'd seen it wasn't the case

acalhounRH:
Sample of data in Grafana dashboard. 

http://marquez.perf.lab.eng.rdu2.redhat.com:3000/d/-8bpuoNWk/fio-summary-with-rook-ceph?orgId=1&from=1571348598365&to=1571348867521&var-user=acalhoun&var-UUID=debug-fio-numjob&var-cluster_name=acalhoun&var-Operation=All&var-io_size=All&var-log_interval=1s&var-net_device=All&var-block_device=All&var-instances=All

acalhounRH:
Sample Grafana data with numjob set to 4. Test was 3 samples or write operations at 4KiB io size with a duration of 15 minutes each, and 32 servers. 

http://marquez.perf.lab.eng.rdu2.redhat.com:3000/d/-8bpuoNWk/fio-summary-with-rook-ceph?orgId=1&from=1571353092644&to=1571356330861&var-user=acalhoun&var-UUID=debug-fio-numjob&var-clustername=acalhoun&var-Operation=All&var-io_size=All&var-interval=30s&var-net_device=All&var-block_device=All&var-instances=All

aakarshg:
Looks like a bad rebase as if you like at file changes it shows a ton more changes that're already in master, can you please fix it @acalhounRH ? and just make sure that the only change noticeable is the one that you made?

fio_wrapper/trigger_fio.py
dustinblack: Can you elaborate on where you see the problem? The line your comment is on is unchanged from current code, as is the entire `try` here except for replacing the `1` with `str(numjob)`, AFAICT.

aakarshg: I'm referring to the timestamp which is the exact line I commented on, the individual log files don't have absolute timestamp, here we are using the start time found in the result json document and assuming it'll be the same for all the jobs( hence the line hasn't changed ). A little concerned about it, but not a blocker if we're all in agreement that this assumption isn't a big problem :)

acalhounRH: I don't believe this is an issue because fio client is the one writing these logs not separate fio instances, I assume the client at X "time past start time" writes down how many iops(or other metric) were produced for each server and job. that would be based on start time and if some servers/jobs were delayed they would have a lower iops count at X "time past start time". at the end of the job the client would then issue a stop command and all instance would die. as of right now the data supports this.

Issue #75, opened 2019-10-18
Failed to account for ramp time when processing logs...


acalhounRH:
Failing to account for ramp time when processing logs. As specified by FIO ramp time is:

> If set, fio will run the specified workload for this amount of time before logging any performance numbers. Useful for letting performance settle before logging results, thus minimizing the runtime required for stable results. Note that the ramp_time is considered lead in time for a job, thus it will increase the total runtime if a special timeout or runtime is specified.

Because of this when ever ramp time is used, the logged start time must be adjusted based on it before processing log results. If this is not fixed and ramp time is used, results will be indexed with incorrect timing. 

if ramp_time:
start time = ramp_time + fio_timestamp
else:
start_time = fio_timestamp

for example see graph a  ramp time of 300s and run time of 600s were used. total iops should be delayed by 300s although it is recorded as starting at the beginning of the testing period. 

![ramp_time bug](https://user-images.githubusercontent.com/31517635/67090820-15c3c880-f179-11e9-86db-aeab396f5c76.png)

PR #78, opened 2019-10-18
fix for issue #75 - failed to account for ramp time

acalhounRH:
see Issue #75 for details of issue

@dustinblack 
@aakarshg

### DIFF
--- a/fio_wrapper/trigger_fio.py
+++ b/fio_wrapper/trigger_fio.py
@@ -57,10 +57,28 @@ class _trigger_fio:
                 document['global_options'] = fio_jobs_dict['global']
             processed.append(document)
             if result['jobname'] != 'All clients':
-                start_time= (int(end_time) * 1000)
-                fio_starttime[result['hostname']] = start_time
+                
+                ramp_time = 0 
+                if 'ramp_time' in result['job options']:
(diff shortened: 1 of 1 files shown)


acalhounRH:
reran testing with fix, data in grafana shows the correct indexing of data, see below.

ramp time was 300s and runtime was 600s, as below you can see that logs have been adjust for the ramp time. 

ramp_time_fixed

fio_wrapper/trigger_fio.py
```diff
@@ -57,10 +57,28 @@ def _document_payload(self, data, user, uuid, sample, list_hosts, end_time, fio_
                 document['global_options'] = fio_jobs_dict['global']
             processed.append(document)
             if result['jobname'] != 'All clients':
-                start_time= (int(end_time) * 1000)
-                fio_starttime[result['hostname']] = start_time
+                
+                ramp_time = 0 
+                if 'ramp_time' in result['job options']:
+                    ramp_time = int(result['job options']['ramp_time'])
+                elif 'ramp_time' in document['global_options']:
```
aakarshg: just a suggestion: this should be an if, or have an else clause as well for code readability, not blocking the pr for this.

PR #80, opened 2019-10-21
Smf per thread results

bengland2:


smallfile_wrapper/trigger_smallfile.py
```diff
@@ -69,18 +69,28 @@ def emit_actions(self):
...
+                    thrd['optype'] = operation
+                    thrd['host'] = self.host
+                    thrd['tid'] = tid
+                    thrd['date'] = timestamp
+                    yield thrd, '-results'
 
             # process response time data
 
-            elapsed_time = float(data['results']['elapsed-time'])
+            elapsed_time = float(data['results']['elapsed'])
```
aakarshg: was this change intended @bengland2 ? confirming before I merge

bengland2:
@aakarshg  yes it was intended.   smallfile itself changed so that the same key 'elapsed' is used on the thread, host, and cluster levels.  'elapsed' generally implies "elapsed time".  In general smallfile output module was rewritten to make JSON output more consistent and K8S friendly (for example, if host-set not specified, then there are only per-thread and cluster results, no host results, since in K8S host=pod).    This PR assumes that behavior now.    When you rebuild quay.io/cloud-bulldozer/snafu it will pull in new version of smallfile.

PR #81, opened 2019-10-21
make it easier to see ES docs when debugging wrapper

bengland2:
set log level to DEBUG if snafu_debug env. var. defined
otherwise set it to INFO

### DIFF
--- a/run_snafu.py
+++ b/run_snafu.py
@@ -35,17 +35,23 @@ es_log.setLevel(logging.CRITICAL)
 urllib3_log = logging.getLogger("urllib3")
 urllib3_log.setLevel(logging.CRITICAL)
 
-setup_loggers("snafu", logging.DEBUG)
 
 def main():
 
     #collect arguments
     parser = argparse.ArgumentParser(description="run script")
     parser.add_argument(
-        '-t', '--tool', action='store', dest='tool', help='Provide tool name')
+        '-v', '--verbose', action='store_const', dest='loglevel', const=logging.DEBUG, default=logging.INFO, help='enables verbose wrapper debugging info')
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


run_snafu.py
```diff
@@ -35,7 +35,12 @@
 urllib3_log = logging.getLogger("urllib3")
 urllib3_log.setLevel(logging.CRITICAL)
 
-setup_loggers("snafu", logging.DEBUG)
+if os.getenv("snafu_debug") != None:
```
aakarshg: Instead of this, can we follow the methodology that most python programs follow i.e. through simple argparse:
```
     parser.add_argument(
        "-v",
        "--verbose",
        dest="loglevel",
        help="set loglevel to INFO",
        action="store_const",
        const=logging.INFO)
    parser.add_argument(
        "-vv",
        "--very-verbose",
        dest="loglevel",
        help="set loglevel to DEBUG",
        action="store_const",
        const=logging.DEBUG)
```

run_snafu.py
```diff
@@ -35,7 +35,12 @@
...
+    setup_loggers("snafu", logging.DEBUG)
+    logger.info("logging level is DEBUG")
+else:
+    setup_loggers("snafu", logging.INFO)
```
aakarshg: and then this can be 

```
setup_loggers("snafu", args.loglevel) 
```

acalhounRH: +1

bengland2: that's ok with me, my point is that there should be SOME way other than code change to have run_snafu.py turn on display of documents that it's pushing to ES.    And I'll add change to documentation to explain how to do this.

bengland2:
The above commit implements aakarsh's suggestion.  OK?

PR #84, opened 2019-10-22
can support non-K8S env, how to debug, additional benchmarks

bengland2:
this documentation PR just explains some things I learned about Snafu that others might be interested in.


bengland2:
@aakarshg I think this is what you wanted.

PR #97, opened 2019-11-20
Adding CI testing

dry923:
Adding CI testing for SNAFU.

Notes:
- The test tries to intelligently run tests based on whats changed. If anything in the root, ci, or utils directory it runs all the test otherwise just the tests in the directories that changed.
- The tests rely on ripsaw so if a test is broken there it will fail the ci here
- Currently hammerdb, smallfile and fs-drift require PR's to be completed in ripsaw for the tests here to Pass
- For tests that index to ES we run a check to see if data is there for the index and uuid.
- Some tests do not have a wrapper so we do a functionality test by simply executing the ripsaw ci test to ensure the image is functioning.

### DIFF
--- a/ci/check_es.py
+++ b/ci/check_es.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import argparse
+import elasticsearch
+import sys
+
+def _check_index(server,port,uuid,index):
+    _es_connection_string = str(server) + ':' + str(port)
+    es = elasticsearch.Elasticsearch([_es_connection_string],send_get_body_as='POST')
+    results = es.search(index=index, body={'query': {'term': {'uuid.keyword': uuid}}}, size=1)
+    if results['hits']['total']['value'] > 0:
+        return 0
+    else:
+        print("No result found in ES")
+        return 1
+
+def main():
+    parser = argparse.ArgumentParser(description="Script to verify uploads to ES")
+    parser.add_argument(
+        '-s', '--server',
+        help='Provide elastic server information')
+    parser.add_argument(
+        '-p', '--port',
+        help='Provide elastic port information')
+    parser.add_argument(
+        '-u', '--uuid',
+        help='UUID to provide to search')
+    parser.add_argument(
+        '-i', '--index',
+        help='Index to provide to search')
+    args = parser.parse_args()
+    
+    sys.exit(_check_index(args.server,args.port,args.uuid,args.index))
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/fio_wrapper/ci_test.sh
+++ b/fio_wrapper/ci_test.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -x
+
+source ci/common.sh
+
+# Build image for ci
+podman build --tag=quay.io/cloud-bulldozer/fio:snafu_ci -f fio_wrapper/Dockerfile . && podman push quay.io/cloud-bulldozer/fio:snafu_ci
+
+cd ripsaw
+
+sed -i 's/fio:latest/fio:snafu_ci/g' roles/fio-distributed/templates/*
+
+# Build new ripsaw image
+update_operator_image snafu_ci
+
+get_uuid test_fiod.sh
+uuid=`cat uuid`
+
+cd ..
+
+# Define index
+index="ripsaw-fio-results ripsaw-fio-log ripsaw-fio-analyzed-result"
+
+check_es $uuid $index
+exit $?
--- a/pgbench-wrapper/ci_test.sh
+++ b/pgbench-wrapper/ci_test.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -x
+
+source ci/common.sh
+
+# Build image for ci
+podman build --tag=quay.io/cloud-bulldozer/pgbench:snafu_ci -f pgbench-wrapper/Dockerfile . && podman push quay.io/cloud-bulldozer/pgbench:snafu_ci
+
+cd ripsaw
+
+sed -i 's/latest/snafu_ci/g' roles/pgbench/defaults/main.yml
+
+# Build new ripsaw image
+update_operator_image snafu_ci
+
+get_uuid test_pgbench.sh
+uuid=`cat uuid`
+
+cd ..
+
+index="ripsaw-pgbench-summary"
+
+check_es $uuid $index
+exit $?
--- a/smallfile_wrapper/ci_test.sh
+++ b/smallfile_wrapper/ci_test.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+set -x
+
+source ci/common.sh
+
+# Build image for ci
+podman build --tag=quay.io/cloud-bulldozer/smallfile:snafu_ci -f smallfile_wrapper/Dockerfile . && podman push quay.io/cloud-bulldozer/smallfile:snafu_ci
+
+cd ripsaw
+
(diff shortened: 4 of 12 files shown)


bengland2:
@dry do fio, smallfile and fs-drift wrappers all fail because they don't have PR 99?   So maybe we need to merge PR 99 first, then PR 97?   But some of the other wrappers are failing too, if absence of PR 99 was to blame, I would expect to see all pass except the above 3, right?

dry923:
This worked for me locally. Trying to debug....
Test | Result | Runtime
-----|--------|--------
fio_wrapper, PASS, 00:06:21
fs_drift_wrapper, FAIL, 00:04:02
hammerdb, FAIL, 00:10:55
iperf, PASS, 00:03:32
pgbench-wrapper, PASS, 00:04:38
smallfile_wrapper, FAIL, 00:03:38
sysbench, PASS, 00:02:44
uperf-wrapper, PASS, 00:06:28
ycsb-wrapper, PASS, 00:09:53

dry923:
ci server was missing elasticsearch module for python. re-running

ycsb-wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+echo $uuid > uuid
+)
+uuid=`cat uuid`
+
+cd ..
+
+# Define ES server info (this is the same from the test cr
+es_server="marquez.perf.lab.eng.rdu2.redhat.com"
+es_port=9200
+index="ripsaw-ycsb-summary-01"
```
aakarshg: this is `ripsaw-ycsb-summary`

pgbench-wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+echo $uuid > uuid
+)
+uuid=`cat uuid`
+
+cd ..
+
+# Define ES server info (this is the same from the test cr
+es_server="marquez.perf.lab.eng.rdu2.redhat.com"
+es_port=9200
+index="ripsaw-pgbench-summary-01"
```
aakarshg: `ripsaw-pgbench-summary`

smallfile_wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+uuid=`cat uuid`
+
+cd ..
+
+# Define ES server info (this is the same from the test cr
+es_server="marquez.perf.lab.eng.rdu2.redhat.com"
+es_port=9200
+index="ripsaw-smallfile-results"
+
+python3 ci/check_es.py -s $es_server -p $es_port -u $uuid -i $index
```
aakarshg: @bengland2 so is it enough to say the code is working fine if we can see documents in `ripsaw-smallfile-results` or should we also check in the `ripsaw-smallfile-rsptimes` ?

fio_wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+echo $uuid > uuid
+)
+uuid=`cat uuid`
+
+cd ..
+
+# Define ES server info (this is the same from the test cr
+es_server="marquez.perf.lab.eng.rdu2.redhat.com"
+es_port=9200
+index="ripsaw-fio-results"
```
aakarshg: we'll need to also look at `ripsaw-fio-logs` and `ripsaw-fio-analyzed_result` for full coverage.

ci/check_es.py
```diff
@@ -0,0 +1,47 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import argparse
+import elasticsearch
+import sys
+
+def _check_index(server,port,uuid,index):
+    es = elasticsearch.Elasticsearch([{'host': server,'port': port}])
+    results = es.search(index=index, body={'query': {'match': {'uuid': {'query': uuid, 'operator': 'AND'}}}})
```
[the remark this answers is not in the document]
aakarshg: Just a suggestion( more like a fyi ) we should add a `size=1` in the search call, we don't really want to look at data ( not setting size would fetch us 10 documents by default ) this will make the query execution faster as well as reduce the load on ES

smallfile_wrapper/ci_test.sh
bengland2: well, it posts to both indices.    But if it got into ripsaw-smallfile-results there is a strong probability that it also got into ripsaw-smallfile-rsptimes.   For now, I'd settle for ripsaw-smallfile-results.

smallfile_wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+set -x
+
+source ci/common.sh
+
+# Build image for ci
+podman build --tag=quay.io/cloud-bulldozer/smallfile:snafu_ci -f smallfile_wrapper/Dockerfile . && podman push quay.io/cloud-bulldozer/smallfile:snafu_ci
+
+cd ripsaw
+
+sed -i 's/smallfile:master/smallfile:snafu_ci/g' roles/smallfile-bench/tempates/*
```
bengland2: typo?  tempates -> templates?

smallfile_wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+
+finish
+echo $uuid > uuid
+)
+uuid=`cat uuid`
+
+cd ..
+
+# Define ES server info (this is the same from the test cr
+es_server="marquez.perf.lab.eng.rdu2.redhat.com"
```
bengland2: why do es_server and es_port get defined inside wrappers instead of a shared environment variable or file?

pgbench-wrapper/ci_test.sh
```diff
@@ -0,0 +1,39 @@
...
+podman build --tag=quay.io/cloud-bulldozer/pgbench:snafu_ci -f pgbench-wrapper/Dockerfile . && podman push quay.io/cloud-bulldozer/pgbench:snafu_ci
+
+cd ripsaw
+
+sed -i 's/latest/snafu_ci/g' roles/pgbench/defaults/main.yml
+
+# Build new ripsaw image
+update_operator_image snafu_ci
+
+sed -i '/trap finish EXIT/d' tests/test_pgbench.sh
```
bengland2: could lines 17-29 get made into a common subroutine shared by all wrappers where you pass the ripsaw script to the subroutine?   This would shrink the CI code considerably.

ci/check_es.py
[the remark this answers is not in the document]
dry923: If I don't use the and it becomes an implicit or and always returns true even if it doesn't find anything.

smallfile_wrapper/ci_test.sh
dry923: updated to check for both

smallfile_wrapper/ci_test.sh
dry923: they now get defined in common.sh which is sourced in.

pgbench-wrapper/ci_test.sh
dry923: moved to get_uuid function. Good call!

bengland2:
Much better IMO, thx.

ci/check_es.py
```diff
@@ -0,0 +1,48 @@
...
+#   limitations under the License.
+
+import argparse
+import elasticsearch
+import sys
+
+def _check_index(server,port,uuid,index):
+    _es_connection_string = str(server) + ':' + str(port)
+    es = elasticsearch.Elasticsearch([_es_connection_string],send_get_body_as='POST')
+    results = es.search(index=index, body={'query': {'match': {'uuid': {'query': uuid, 'operator': 'AND'}}}}, size=1)
```
aakarshg: There are 2 fields for uuid, uuid itself which is of type text so analyzers are used and then the keyword subfield which is of type keyword which we can use a term query on. The mapping for uuid looks as follows:
```
        "uuid" : {
          "type" : "text",
          "fields" : {
            "keyword" : {
              "type" : "keyword",
              "ignore_above" : 256
            }
          }
        }
```
So the reason why you're forced to use an AND is because we're using a match query on the text field, instead if we use the [term query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-term-query.html) on the keyword field it'll look return documents where uuid is set to the value we gave. 

So the query should look like: 
```
{
    "query": {
        "term": {
            "uuid.keyword": uuid
        }
    }
}
```
HTH, the reason why we don't want to do the match query is because it'll search for the input uuid in uuid of documents so if the uuid is 1234, then it'll also return documents with uuid 12345. but in reality we only want documents with uuid set to 1234. This is just an edge case but hope it gives an idea of why we should use the term query in this case ( also it's faster because it creates the lookups ).

PR #107, opened 2019-12-20
Skip not present CI tests

### LINKED ISSUE #112: no cluster_loader/ci_test.sh
causes CI to fail.   For example, see [this log error](http://perf-sm5039-4-8.perf.lab.eng.rdu2.redhat.com:8080/job/snafu_jjb/25/consoleFull)

rsevilla87:
Non existing tests, like cluster loader, breaks CI. Skip them if they doesn't exist.

Closes #112

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -43,21 +43,23 @@ test_rc=0
 
 for dir in `echo $test_list`
 do
-  start_time=`date`
   my_dir=${dir::-1}
-  figlet "CI test for "$my_dir
-  if $my_dir/ci_test.sh
-  then
-    end_time=`date`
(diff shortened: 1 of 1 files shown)


aakarshg:
ok to test

ci/run_ci.sh
```diff
@@ -46,7 +46,7 @@ do
   start_time=`date`
   my_dir=${dir::-1}
   figlet "CI test for "$my_dir
-  if $my_dir/ci_test.sh
+  if [ -f $my_dir/ci_test.sh ] && $my_dir/ci_test.sh
```
aakarshg: just this change alone will still cause failures in CI as the rc is set in else part. Unless im reading this wron

rsevilla87: You're right! I pushed some changes fixing this

PR #109, opened 2020-01-02
Improve logic with checking for es

aakarshg:
Changes to snafu, to only index if the es env variables
aren't empty strings.

### DIFF
--- a/run_snafu.py
+++ b/run_snafu.py
@@ -58,9 +58,16 @@ def main():
     #instantiate elasticsearch instance and check connection
     es={}
     if "es" in os.environ:
-        es['server'] = os.environ["es"]
-        es['port'] = os.environ["es_port"]
-        index_args.prefix = os.environ["es_index"]
+        if os.environ["es"] != ""
+            es['server'] = os.environ["es"]
+            logger.info("Using elasticsearch server with host:" + str(es['server']))
+        if os.environ["es_port"] != ""
+            es['port'] = os.environ["es_port"]
+            logger.info("Using elasticsearch server with port:" + str(es['port']))
+    if len(es.keys()) == 2:
+        if os.environ["es_index"] != "":
+            index_args.prefix = os.environ["es_index"]
+            logger.info("Using index prefix for ES:" + str(index_args.prefix))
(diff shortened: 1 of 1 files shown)


aakarshg:
@chaitanyaer PTAL

run_snafu.py
```diff
@@ -58,9 +58,13 @@ def main():
...
-        es['server'] = os.environ["es"]
-        es['port'] = os.environ["es_port"]
-        index_args.prefix = os.environ["es_index"]
+        if os.environ["es"] != ""
+            es['server'] = os.environ["es"]
+        if os.environ["es_port"] != ""
+            es['port'] = os.environ["es_port"]
+    if len(es.keys()) == 2:
+        if os.environ["es_index"] != "":
+            index_args.prefix = os.environ["es_index"]
```
jtaleric: Good idea. Can we also throw out a log message? Something that lets the user know what server/port/index was used? (Yes they should know, but is always good to confirm)

aakarshg: L72 will already be giving a full set of details, but updated the PR so that we're also showing info about what's server/port/index.

run_snafu.py
```diff
@@ -58,9 +58,16 @@ def main():
...
     if "es" in os.environ:
-        es['server'] = os.environ["es"]
-        es['port'] = os.environ["es_port"]
-        index_args.prefix = os.environ["es_index"]
+        if os.environ["es"] != ""
+            es['server'] = os.environ["es"]
+            logger.info("Using elasticsearch server with host:" + str(es['server']))
+        if os.environ["es_port"] != ""
+            es['port'] = os.environ["es_port"]
+            logger.info("Using elasticsearch server with port:" + str(es['port']))
```
chaitanyaenr: Can we also add a log message about the defaults when the es details are empty?

aakarshg: We're not using any defaults with snafu, so if the user passes any vars as env then we'll show it otherwise there's no values.

jtaleric: Yeah, something to the effect of `not indexing results for long term storage`

PR #113, opened 2020-01-09
Improve image build & push reliability

rsevilla87:
One of the issues we had, is that on failed builds the CI system keeps going and pushes an old cached image.
The same happens for push operations.
The new function `build_and_push` takes care exiting on failed image builds & push. It also includes a 3 retries logic for the image push operation.


aakarshg:
ok to test

rsevilla87:
\o/

PR #116, opened 2020-01-14
add support for parallel cl

aakarshg:
This provides user with the option to use the parallel version if needed ( case of mastervert ).


aakarshg:
Closing this until we can come together on which cluster loader to use and which invocation to rely on.

PR #118, opened 2020-01-14
Fixing ambiguous master error

dry923:
CI is getting the following error when doing a git diff. As a result it is always running against all the tests. This is to resolve it.

```
...
++ git diff master --name-only
fatal: ambiguous argument 'master': unknown revision or path not in the working tree.
...
```


dry923:
This now works appropriately. From CI:
```
...
++ git diff origin/master --name-only
+ diff_list=ci/run_ci.sh
...
```

PR #120, opened 2020-01-15
Enhancement for HammerDB doc generation/indexing

mkarg75:
hopefully the last one


bengland2:
@mkarg75 , I'm concerned that the results you get from minishift might not match the results other people get from OCP4 or from minikube.  Does that make sense?  If so, any chance you can try this with version of minikube recommended by Russ or whoever in outcome of  [ripsaw issue  272](https://github.com/cloud-bulldozer/ripsaw/issues/272)  ?

bengland2:
also would like to see this patch in snafu/hammerdb/hammerd_wrapper.py:
```
@@ -170,6 +170,8 @@ def main():
     if es_server != "" :
         if len(documents) > 0 :
             _index_result("ripsaw-hammerdb-results", es_server, es_port, documents)
+        else:
+            raise Exception('failed to produce hammerdb results document')
```
or something equivalent to it, where benchmark doesn't just silently fail.

hammerdb/Dockerfile
```diff
@@ -2,30 +2,40 @@ FROM centos/tools
...
 
+RUN yum install -y git python-pip
 RUN pip install --upgrade pip
 RUN pip install "elasticsearch>=6.0.0,<=7.0.2"
 
-RUN mkdir -p /opt/snafu
-COPY . /opt/snafu
+# clone the snafu repo 
+#RUN cd /opt/; git clone https://github.com/mkarg75/snafu.git
+RUN cd /opt/; git clone https://github.com/cloud-bulldozer/snafu.git
```
aakarshg: This change is not required, as snafu is in the build context of the image.

dry923:
@mkarg75 looks like hammer was successful however it failed ci because it couldn't find any results in ES.

```
++ kubectl logs hammerdb-workload-723e8b18-j89x8 --namespace my-ripsaw
++ grep 'SEQUENCE COMPLETE'
++ echo 'Hammerdb test: Success'
Hammerdb test: Success
...
+ python3 ci/check_es.py -s marquez.perf.lab.eng.rdu2.redhat.com -p 9200 -u 723e8b18-644f-5352-aafe-c18c679ca138 -i ripsaw-hammerdb-results
No result found in ES
```

dry923:
@mkarg75 why the revert from the ubi8 Dockerfile?

hammerdb/build_container.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
aakarshg: can you please remove this file?

hammerdb/entrypoint
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
```
aakarshg: Wouldn't this file make uid_entrypoint file redundant ? in that case can we remove one from the other ?

aakarshg: I don't see this added to the dockerfile as well?

hammerdb/hammerd_wrapper.py
```diff
@@ -34,59 +34,69 @@ def _parse_stdout(stdout):
...
-        print("Timed test: {}".format(entry['timed_test']))
-        print("HammerDB results (TPM):")
-        print("""
-              TPM: {}""".format(entry['tpm']))
-        print("HammerDB results (NOPM):")
-        print("""
-              NOPM: {}""".format(entry['nopm']))
-        print("Timestamp: {}".format(entry['timestamp']))
-        print("+{}+".format("-"*(115)))
+    max_workers = int(data[0]['db_num_workers'])
```
aakarshg: We already know the number of workers and samples through the variables db_num_workers and samples so can we just use that here instead of getting it through the stdout again ?

mkarg75: we don't know them at this point, since they are not global but we need them to iterate properly through the outer loop of workers and the inner loop of samples.

PR #121, opened 2020-01-16
fix bug trimming last char in test dir

aakarshg:
Without this ci would fail on trying to run against the workload (-last char ) so for hammerdb it'd try to look into hammerd instead of hammerdb.

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -43,7 +43,7 @@ test_rc=0
 
 for dir in `echo $test_list`
 do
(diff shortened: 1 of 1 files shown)


aakarshg:
Fio, fs-drift, hammerdb, pgbench, uperf and smallfile all seem to be breaking due to the same error:
```
Error copying image to the remote destination: Error writing blob: Patch
```
by doing a find with above string on the ci logs you can notice it.

PR #123, opened 2020-01-16
CI code improvements

rsevilla87:
This PR aims to improve ci code with things such as simplify some operatives and use more reliable methods to look for available ci scripts.
Helps on https://github.com/cloud-bulldozer/snafu/issues/111

### DIFF
--- a/ci/run_ci.sh
+++ b/ci/run_ci.sh
@@ -10,7 +10,7 @@ kubectl create namespace my-ripsaw
 
 # Clone ripsaw so we can use it for testing
 rm -rf ripsaw
-git clone https://github.com/cloud-bulldozer/ripsaw.git
+git clone https://github.com/cloud-bulldozer/ripsaw.git --depth 1
 
 if [[ $ghprbPullLongDescription = *"Depends-On:"* ]]; then
   ripsaw_change_id="$(echo $ghprbPullLongDescription | sed -n -e 's/^.*Depends-On: //p')"
@@ -23,10 +23,12 @@ fi
 
 
 # Prep results.markdown file
-echo "Results for SNAFU CI Test" > results.markdown
-echo "" >> results.markdown
-echo 'Test | Result | Runtime' >> results.markdown
-echo '-----|--------|--------' >> results.markdown
+cat > results.markdown << EOF
+Results for SNAFU CI Test
+
+Test | Result | Runtime
+-----|--------|--------
+EOF
 
 diff_list=`git diff origin/master --name-only`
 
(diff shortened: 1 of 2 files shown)


ci/run_ci.sh
```diff
@@ -28,38 +30,32 @@ diff_list=`git diff origin/master --name-only`
 # - anything in ci has been changed
 # - anything in utils has been changed
 # Else only run tests on directories that have changed
-if [[ `echo $diff_list | grep -v / | wc -l` -gt 0 || `echo $diff_list | grep ci/` || `echo $diff_list | grep utils/` ]]
-then
+if [[ `echo ${diff_list} | grep -cv /` -gt 0 || `echo ${diff_list} | grep -E "(ci|utils|image_resources)/"` ]]; then
   echo "Running full test"
-  test_list=`ls -d */ | grep -Ev "(utils|ci|ripsaw|image_resources)/"`
+  test_list=`find * -name ci_test.sh -type f -exec dirname {} \;`
```
bengland2: would -maxdepth 1 be appropriate here because the ci_test.sh has to appear in the top wrapper directory?   just being paranoid, but that matches what ls -d */ does more closely.  But overall this is an improvement, gets rid of grep -v.

ci/run_ci.sh
```diff
@@ -28,38 +30,32 @@ diff_list=`git diff origin/master --name-only`
...
 else
   echo "Running specific tests"
   echo $diff_list
-  test_list=`echo $diff_list | awk -F "/" '{print $1}' | uniq`
+  test_list=`echo "${diff_list}" | xargs dirname | uniq`
```
bengland2: call me paranoid, but if someone creates a wrapper with subdirectories, would this work?   I think the awk command would.    I don't think any existing wrappers have subdirectories, but that could change, there is nothing requiring a wrapper not to use subdirectories.   I don't think this is an essential change, do you?

ci/run_ci.sh
```diff
@@ -28,38 +30,32 @@ diff_list=`git diff origin/master --name-only`
...
-      duration=`date -ud@$(($(date -ud"$end_time" +%s)-$(date -ud"$start_time" +%s))) +%T`
-      echo ${dir::-1}" | PASS | "$duration >> results.markdown
-    else
-      end_time=`date`
-      duration=`date -ud@$(($(date -ud"$end_time" +%s)-$(date -ud"$start_time" +%s))) +%T`
-      echo ${dir::-1}" | FAIL | "$duration >> results.markdown
-      test_rc=1
-    fi
-    wait_clean
+for dir in ${test_list}; do
```
bengland2: so if I understand this correctly, we got rid of the -f $my_dir/ci_test.sh because the test_list was constructed to have only directories containing a ci_test.sh.   The only other change I see here is ${dir::-1} to ${dir}, which makes sense because if "dir" is an empty string or undefined then how did it get into test_list?    it's hard to see if there are other changes here because of the indentation change, but I think that's all, right?

ci/run_ci.sh
rsevilla87: You're right, this change is not essential. It is just an attempt to get the rid of awk here, which sometimes can be confusing, as this command is not focused on path parsing but in strings processing. You're also right in the fact that we're not currently using sub-directories, dirname might be problematic here.

ci/run_ci.sh
rsevilla87: Adding -maxdepth flag... Thanks for the suggestion!

bengland2:
LGTM now  -ben

ci/run_ci.sh
rsevilla87: Yes, this is exactly what I wanted to do.

dry923:
@rsevilla87 FYI theres a conflict in ci/common.sh

bengland2:
I don't understand output of snafu_linters CI job.   It seems not to have run at all, did it?
```
> git rev-parse origin-pull/pull/130/merge^{commit} # timeout=10
ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
```
@aakarshg

dry923:
> I don't understand output of snafu_linters CI job. It seems not to have run at all, did it?
> 
> ```
> > git rev-parse origin-pull/pull/130/merge^{commit} # timeout=10
> ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
> ```
> 
> @aakarshg

@aakarshg hey any idea on ^

aakarshg:
> > I don't understand output of snafu_linters CI job. It seems not to have run at all, did it?
> > ```
> > > git rev-parse origin-pull/pull/130/merge^{commit} # timeout=10
> > ERROR: Couldn't find any revision to build. Verify the repository and branch configuration for this job.
> > ```
> > 
> > 
> > @aakarshg
> 
> @aakarshg hey any idea on ^

this was when we're testing the job. It'll run fine now if we hit the rerun

PR #126, opened 2020-01-19
Adding the vdbench benchmark

Avilir:
Signed-off-by: Avi Liani <alayani@redhat.com>


Avilir:
@bengland2 , there is no option to pull the vdbench from the oracle site via CLI, since you need the sign the ALUA before you can download it. this is the reason i add the binary files that it need.
About other files that can be removed, i will recheck and removed them.

Avilir:
I removed all binary files from the repository.

vdbench_wrapper/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM quay.io/avili/vdbench:1.0.1
+
+USER root
```
bengland2: don't think this line is needed, I don't use it.

vdbench_wrapper/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+RUN dnf install --nodocs -y python3-numpy
+RUN dnf clean all
```
bengland2: not clear why this is needed either.   All that does is delete caches right?  We want to exploit the cache for reasons you'll soon discover ;-)

vdbench_wrapper/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN pip3 install --user elasticsearch configparser statistics requests
+RUN pip3 install --upgrade-strategy=only-if-needed pyyaml "elasticsearch>=6.0.0,<=7.0.2"
+
+RUN  mkdir -p /opt/snafu/
+COPY . /opt/snafu/
+
+WORKDIR /opt/snafu
+CMD ["/bin/bash"]
```
bengland2: is this needed?

vdbench_wrapper/ci_test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -x
```
bengland2: Raul taught me a bash trick to prevent command failures from being ignored:
```
set -xeEo pipefail
```
Highly recommend.   This avoids a lot of cases where benchmark fails but CI reports success or fails much later.

vdbench_wrapper/vdbench_wrapper.py
```diff
@@ -0,0 +1,126 @@
...
+        # collect arguments
+        parser.add_argument(
+            '-c', '--config',
+            help='Provide configuration file')
+        parser.add_argument(
+            '-o', '--output',
+            help='path to output files',
+            default=BASE_PATH+'/outputs')
+        parser.add_argument(
+            '-s', '--sample',
```
bengland2: nitpicking for consistency with other benchmarks -- "--sample" => "--samples" ?

vdbench_wrapper/vdbench_wrapper.py
```diff
@@ -0,0 +1,126 @@
...
+class vdbench_wrapper():
+
+    def __init__(self, parser):
+
+        # collect arguments
+        parser.add_argument(
+            '-c', '--config',
+            help='Provide configuration file')
+        parser.add_argument(
+            '-o', '--output',
```
bengland2: nitpick for consistency with other benchmarks - change to '-d', '--dir'   .   Also, in most utilities -o usually means output file (singular), this parameter is a directory is it not?

vdbench_wrapper/vdbench_wrapper.py
```diff
@@ -0,0 +1,126 @@
...
+            '-s', '--sample',
+            help='number of times to run benchmark, defaults to 1',
+            type=int,
+            default=1)
+        parser.add_argument(
+            '-l', '--log',
+            help='log file (Full path)',
+            default=BASE_PATH+'/logs')
+        parser.add_argument(
+            '-d', '--dir',
```
bengland2: why do you need this if you have output dir above?

bengland2:
oh one other comment from Joe T - why not use byowl (bring your own workload) benchmark?  I have to look into what he meant by this but basically it is there as a catch all for workloads that can't be built into container images by snafu.

vdbench_wrapper/Dockerfile
Avilir: I delete it.

vdbench_wrapper/Dockerfile
Avilir: Delete it.

vdbench_wrapper/ci_test.sh
Avilir: Changed it.

vdbench_wrapper/vdbench_wrapper.py
Avilir: Changed it.

vdbench_wrapper/vdbench_wrapper.py
Avilir: The purpose of the directorys is different.
The first one (output dir) is used for the benchmark output itself for each test. and this one is where to put the Results.tgz file which contain all outputs, logs and reports from all samples.

vdbench_wrapper/Dockerfile
Avilir: I delete it.

PR #127, opened 2020-01-20
Fixed printing results for hammerdb

SzymonZebrowski:
During printing results of benchmark, for each test only results of first test were printed


aakarshg:
ok to test

PR #133, opened 2020-02-03
[WIP] Add makefile

rsevilla87:
Add makefile as wrapper to build, push or clean containers and test Snafu code.

An example of the makefile's help
```
$ make help
Targets for snafu:

    make                   Build and push all container images.
    make build             Build all container images.
    make push              Push all container images.
    make clean             Clean all container images.
    make test              Perform CI tests.
    make BENCHMARK         Build and push BENCHMARK container image.
    make build-BENCHMARK   Build BENCHMARK container image.
    make push-BENCHMARK    Build BENCHMARK container image.
    make test-BENCHMARK    Execute BENCHMARK tests.
    make clean-BENCHMARK   Clean BENCHMARK container image.
    make help              Show this message.

Where BENCHMARK may be: fio fs-drift hammerdb iperf pgbench smallfile sysbench uperf ycsb 
```

- The flag `--pull-always` is passed to the build command to ensure using the latest base image available, and avoid using outdated layers and RPMs in later dockerfile steps.
- `SUDO=yes` env. var may be used to execute podman commands with sudo.


bengland2:
suggestion: @rsevilla87 , document it in snafu README.md , even if only to mention "make help" and that every benchmark supports 4 operations, build, push, test and clean, and that if you just use benchmark name as target then that means just build and push, not test.

I like that we are making the benchmark UI look more similar and require less typing, this makes it easier for people to learn how to use snafu.    But the way you are using make is not exactly what people typically do with it....

Don't existing ci_test.sh scripts all do build and push also?   So test-BENCHMARK is actually doing build, push and test, which is counter-intuitive.    What a makefile typically does -- determine if steps are needed, using timestamps from output artifacts for previous steps.   For example, 
I would have guessed that "make test-smallfile" would check if rebuild of smallfile image is necessary, if so  then build and push it before running the test.  But to do that, you'd have to allow "make" to detect if timestamp of image sources was newer than timestamp of the image (available with podman 
 and skopeo).    Perhaps we could have a Makefile for each wrapper, and have it include ../common.make with some common elements like build_container and push_container, but then you'd have a rule in the per-wrapper Makefile that would decide whether to update the wrapper image locally and then push it to the image repo.  Once it's pushed to the image repo, K8S will know to pull down the new image for the test.  But I haven't put much thought into how to actually implement such a Makefile, since "make" isn't aware of file timestamps that are anywhere besides a mountpoint.  I'm just saying that is typically what Makefile files are used for.  A simple shell/python script could do what you are doing with the Makefile today, right?  And with less dependencies (you don't need make RPM).   What do you think?

rsevilla87:
@bengland2 I added a mechanism to avoid rebuilding containers without changes on snafu's code.
I still have pending to make modifications to ci tests to avoid rebuilding containers.

jtaleric:
Checking in on this one?

Issue #111, opened 2020-01-03
CI reliability


rsevilla87:
The following https://github.com/cloud-bulldozer/snafu/pull/109 had syntax errors which weren't detected by CI (fixed by https://github.com/cloud-bulldozer/snafu/pull/110). CI needs some love as it doesn't look very reliable in certain scenarios.

aakarshg:
@dry923 might be worth the time to investigate the issue when you get back and have some cycles.

aakarshg:
@bengland2 has pointed out that the CI hasn't been rebuilding the workload images ( snafu_ci tag ) in over a month, which is evident through logs so this would mean that the CI has been using old images for workloads and thus the tests could have been passing ?

aakarshg:
The fix for this would be to 

> clear out the prior incarnation of the image before pushing the new one?  That way, either the test fails or the image used is actually reflecting the source code that we want to test.

as pointed out by @bengland2

bengland2:
Here is a set of changes that I believe fixes most of the problem.    It isn't fully tested yet because I have to finish other PRs but wanted to make you aware of what I had so far:

https://github.com/bengland2/snafu/tree/ensure-benchmark-updated

I'll submit as a PR if there is interest, but was waiting to get Russ input, etc.
The basic ideas are: 1) centralize as much logic as possible in ci/common.sh, and 2) never assume that a bash command succeeds.   While bash "set -e"  is a brute force way to accomplish this, it also prevents you from doing things like retrying podman push, or using an error status as input to an if conditional.

rsevilla87:
@bengland2 just merged this PR some hours ago. https://github.com/cloud-bulldozer/snafu/pull/113

bengland2:
I didn't know if anyone was working on it, that's why I didn't submit as a PR.  As long as the problem gets solved...

aakarshg:
There's still quite a few issues with CI, if you look at http://perf-sm5039-4-8.perf.lab.eng.rdu2.redhat.com:8080/job/snafu_jjb/87/consoleFull for fs-drift you can see that the uuid cannot even be parsed ( the reason is because we delete all before we can look at the uuid ), but the CI still replied with a passed on fs-drift !

dry923:
[...]

PR #135, opened 2020-02-04
Verify number of arguments on check_es

rsevilla87:
If for any reason `get_uuid` doesn't return any uuid, `check_es` is called with an empty $uuid argument. Due to this behavior, the current implementation of the function can led to false positives during CI tests.
This simple code improvement avoids this.
#111

### DIFF
--- a/fs_drift_wrapper/ci_test.sh
+++ b/fs_drift_wrapper/ci_test.sh
@@ -22,6 +22,5 @@ uuid=`cat uuid`
 cd ..
 
 indexes="ripsaw-fs-drift-results ripsaw-fs-drift-rsptimes ripsaw-fs-drift-rates-over-time"
-for i in $indexes ; do
-  check_es $uuid $i || exit 1
-done
+check_es "${uuid}" "${indexes}"
+exit $?
--- a/README.md
+++ b/README.md
@@ -208,9 +208,9 @@ uuid=`cat uuid`
 
 cd ..
 
-# Define index (there can be more than 1)
+# Define index (there can be more than 1 separated by whitespaces)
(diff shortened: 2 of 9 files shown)


dry923:
@rsevilla87 the function is meant to be able to take 2+ arguments with everything after the uuid being indexes to check. So this would need to be a >=

rsevilla87:
> @rsevilla87 the function is meant to be able to take 2+ arguments with everything after the uuid being indexes to check. So this would need to be a >=

Ooops, you're right, I though it was a comma separated list

dry923:
The other thought is that if you have an instance with 2+ indexes and no uuid it will try and use an index as the uuid. Not sure what the outcome would be there.

rsevilla87:
To avoid that, we could call `check_es ` by quoting the input arguments. e.g.
```
$ f(){ 
  echo ${#}; 
  echo $1; 
  for i in ${2}; do 
    echo $i; 
  done; 
}
$ f "uuid" "i1 i2 i3"
2
uuid
i1
i2
i3
```

aakarshg:
@rsevilla87 can you please rebase this and fix the merge conflicts? thank you.

fs_drift_wrapper/ci_test.sh
```diff
@@ -23,5 +23,5 @@ cd ..
 
 indexes="ripsaw-fs-drift-results ripsaw-fs-drift-rsptimes ripsaw-fs-drift-rates-over-time"
 for i in $indexes ; do
```
dry923: I don't think this for loop is needed given the check_es function loops over the indexes

PR #136, opened 2020-02-05
Add a way for ripsaw to query redis the uperf-test status code

mohit-sheth:
@jtaleric @aakarshg  
With this commit the snafu-uperf will be able to send 3 params to redis (hash_name, test_name and status_code). With this information ripsaw-uperf can not only query to check if there is a failure but it can also identify **which** test/s failed. This can be used to only retry the failed test/s instead of starting from the beginning.  

```The example output shows uperf-test with 2 sets of parameters
10.131.0.143:6379> hget uperf-test uperf-stream-tcp-16384-1
"0"   #Completed
10.131.0.143:6379> hget uperf-test uperf-stream-tcp-16384-8
"1"   #Failed 
10.131.0.143:6379> hgetall uperf-test
1) "uperf-stream-tcp-16384-8"
2) "1"
3) "uperf-stream-tcp-16384-1"
4) "0"
10.131.0.143:6379>
```
This commit also adds the ability to echo the test which failed for better readability 
```
UPerf failed to execute with args uperf-stream-tcp-16384-8, trying one more time..  #from snafu
UPerf failed to execute with args uperf-stream-tcp-16384-8 second time, stopping.. #from snafu
Workload failed with argument stream-tcp-16384-8                    #from ripsaw   
```

### DIFF
--- a/uperf-wrapper/redis_utils.py
+++ b/uperf-wrapper/redis_utils.py
@@ -1,0 +1,12 @@
+#!/usr/bin/python
+
+import redis
+
+class RedisUtils:
+    def __init__(self, redis_host, redis_port):
+        self.redis_host = redis_host
+        self.redis_port = redis_port
+        self.redis_cache = redis.Redis(host=self.redis_host, port=self.redis_port)
+
+    def set_code(self, test_name, test_args, test_status_code,):
+        self.redis_cache.hset(test_name, test_args, test_status_code)
--- a/uperf-wrapper/uperf-wrapper.py
+++ b/uperf-wrapper/uperf-wrapper.py
@@ -191,6 +191,8 @@ def main():
     server_node = ""
     client_node = ""
     args.cluster_name = "mycluster"
+    redis_host = ""
+    redis_port = ""
     if "clustername" in os.environ:
         args.cluster_name = os.environ["clustername"]
     if "serviceip" in os.environ:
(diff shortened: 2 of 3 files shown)


uperf-wrapper/uperf-wrapper.py
```diff
@@ -206,13 +208,20 @@ def main():
...
+        redis_host = os.environ["bo_ip"]
+
+    redis_port = 6379  
+    r = RedisUtils(redis_host, redis_port)
+    r.set_code(args.workload[0].split('/')[-2], args.workload[0].split('/')[-1], "2")
 
     stdout = _run_uperf(args.workload[0])
     if stdout[1] == 1 :
-        print("UPerf failed to execute, trying one more time..")
+        print(f"UPerf failed to execute with args {args.workload[0].split('/')[-1]}, trying one more time..")
```
aakarshg: typo maybe ? 'f'  ?

uperf-wrapper/uperf-wrapper.py
```diff
@@ -206,13 +208,20 @@ def main():
         server_node = os.environ["server_node"]
     if "client_node" in os.environ:
         client_node = os.environ["client_node"]
+    if "bo_ip" in os.environ:
+        redis_host = os.environ["bo_ip"]
+
+    redis_port = 6379
```
aakarshg: make this redis part of code optional, so that if folks are running outside of ripsaw, they should still be able to run with this.

uperf-wrapper/uperf-wrapper.py
mohit-sheth: shorthand for .format("string")

uperf-wrapper/redis_utils.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/python
```
rsevilla87: In my opinion, this code shouldn't be tied to uperf-wrapper, should be somewhere else like the utils module

mohit-sheth: yes, Initially i did put it in utils but was having issues importing the file. I also created a __init__.py there as you can see in the commit.  I will check it out again. Thanks for the feedback

rsevilla87: We're now using python3, so imports must be absolute. i.e
`from utils.redis.utils import RedisUtils`

uperf-wrapper/uperf-wrapper.py
jtaleric: Let's accept the `redis_port` as a param, similar to above.

aakarshg:
@mohit-sheth requires a rebase.

aakarshg:
@mohit-sheth the linters job failed, there's a couple of minor linting errors.

mohit-sheth:
closing in favor of https://github.com/cloud-bulldozer/snafu/pull/144

PR #137, opened 2020-02-05
Re-enabling metadata collection

dry923:
Metadata collection was previously disabled. However, since the ripsaw test waits for it to complete anyway we are waiting longer for the timeout than if the collection was actually done.


rsevilla87:
Nice one. Everything speeding up CI is very well received!
lgtm

PR #138, opened 2020-02-13
added uperf to run from run_snafu --tool

amitsagtani97:
Depends-On: 283


uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,185 @@
...
+                                       args.server_node,
+                                       args.client_node)
+        if args.server != "":
+            if len(documents) > 0:
+                self._index_result("ripsaw-uperf-results", args.server, args.port, documents)
+        print(stdout[0])
+        if len(documents) > 0:
+            self._summarize_data(documents)
+
+    def _index_result(self, index, server, port, payload):
```
aakarshg: this won't be required anymore as we do indexing through the run_snafu

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,185 @@
...
+        data = self._parse_stdout(stdout[0])
+        documents = self._json_payload(data, args.run[0], args.uuid, args.user, args.hostnetwork,
+                                       args.serviceip,
+                                       args.remoteip,
+                                       args.clientips, args.cluster_name, args.resourcetype[0],
+                                       args.server_node,
+                                       args.client_node)
+        if args.server != "":
+            if len(documents) > 0:
+                self._index_result("ripsaw-uperf-results", args.server, args.port, documents)
```
aakarshg: this will need to be `yield documents, 'results' `

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,185 @@
...
+import subprocess
+from datetime import datetime
+
+import elasticsearch
+import numpy as np
+
+
+class Trigger_uperf():
+    def __init__(self, args):
+        stdout = self._run_uperf(args.workload[0])
```
aakarshg: I think all of this will need to move to a new function called emit_actions(), you can take a look at https://github.com/cloud-bulldozer/snafu/blob/master/run_snafu.py#L119 and that's why the comment below about yield documents and the index...

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,194 @@
...
+                print("UPerf failed to execute a second time, stopping...")
+                exit(1)
+        data = self._parse_stdout(stdout[0])
+        documents = self._json_payload(data, self.run[0], self.uuid, self.user, self.hostnetwork,
+                                       self.serviceip,
+                                       self.remoteip,
+                                       self.clientips, self.cluster_name, self.resourcetype[0],
+                                       self.server_node,
+                                       self.client_node)
+        if self.server != "":
```
aakarshg: this is not required...

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,194 @@
...
+import subprocess
+from datetime import datetime
+
+import elasticsearch
+import numpy as np
+
+
+class Trigger_uperf():
+    def __init__(self, args):
+        self.server = args.server
```
aakarshg: this is not required, as the indexing will be done through the main code

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,194 @@
...
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import re
+import subprocess
+from datetime import datetime
+
+import elasticsearch
```
aakarshg: safe to remove this

uperf_wrapper/uperf_wrapper.py
```diff
@@ -0,0 +1,67 @@
...
+        self.args.serviceip = ""
+        self.args.server_node = ""
+        self.args.client_node = ""
+        self.args.cluster_name = "mycluster"
+
+        if "clustername" in os.environ:
+            self.args.cluster_name = os.environ["clustername"]
+        if "serviceip" in os.environ:
+            self.args.serviceip = os.environ['serviceip']
+        if "es" in os.environ:
```
aakarshg: This part of code will be handled by the main code in run_snafu so we can remove this

uperf_wrapper/uperf_wrapper.py
```diff
@@ -0,0 +1,67 @@
...
+            help='Provide XML workload location')
+        parser.add_argument(
+            '-r', '--run', nargs=1,
+            help='Provide the iteration for the run')
+        parser.add_argument(
+            '--resourcetype', nargs=1,
+            help='Provide the resource type for this uperf run - pod/vm/baremetal')
+        self.args = parser.parse_args()
+
+        self.args.server = ""
```
aakarshg: This is not needed here too as this part of logic will be done through the run_snafu logic.

uperf_wrapper/uperf_wrapper.py
amitsagtani97: We will need to update uuid from it.

aakarshg: uuid ? from es vars? I'm not sure I follow

uperf_wrapper/uperf_wrapper.py
```diff
@@ -0,0 +1,65 @@
...
+        self.args.client_node = ""
+        self.args.cluster_name = "mycluster"
+
+        if "clustername" in os.environ:
+            self.args.cluster_name = os.environ["clustername"]
+        if "serviceip" in os.environ:
+            self.args.serviceip = os.environ['serviceip']
+        if "es" in os.environ:
+            self.args.port = os.environ["es_port"]
+            self.args.uuid = os.environ["uuid"]
```
aakarshg: This can be a check of its own, as uuid is not tied to es vars

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,191 @@
...
+                exit(1)
+        data = self._parse_stdout(stdout[0])
+        documents = self._json_payload(data, self.run[0], self.uuid, self.user, self.hostnetwork,
+                                       self.serviceip,
+                                       self.remoteip,
+                                       self.clientips, self.cluster_name, self.resourcetype[0],
+                                       self.server_node,
+                                       self.client_node)
+        if len(documents) > 0:
+            yield documents, 'results'
```
aakarshg: this needs to be:
```
            for document in documents:
                yield document, 'results'
```
We've to yield the individual document and not the list.

aakarshg:
Can you please rebase this @amitsagtani97

aakarshg:
This needs a rebase as well, please rebase

uperf_wrapper/uperf_wrapper.py
```diff
@@ -0,0 +1,62 @@
...
+            help='Provide XML workload location')
+        parser.add_argument(
+            '-r', '--run', nargs=1,
+            help='Provide the iteration for the run')
+        parser.add_argument(
+            '--resourcetype', nargs=1,
+            help='Provide the resource type for this uperf run - pod/vm/baremetal')
+        self.args = parser.parse_args()
+
+        self.args.uuid = ""
```
aakarshg: uuid is set here to an empty string, but not updated wth the uuid from environment variable

uperf_wrapper/uperf_wrapper.py
```diff
@@ -0,0 +1,62 @@
...
+        self.args.user = ""
+        self.args.clientips = ""
```
aakarshg: clientips is set to emoty string but not updated..

uperf_wrapper/trigger_uperf.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def _run_uperf(self, workload):
+        cmd = "uperf -v -a -x -i 1 -m {}".format(workload)
+        process = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE)
+        stdout, stderr = process.communicate()
+        return stdout.strip().decode("utf-8"), process.returncode
+
+    def _parse_stdout(self, stdout):
+        # This will effectivly give us:
+        # ripsaw-test-stream-udp-16384
```
jtaleric: Small nit against myself here... but we don't have the nthr in the output here, but in the code we now have nthr :D

jtaleric:
> The documents in elasticsearch are missing user and client_ips field

Looking at the code, I don't see how this can be? Unless uperf failed to run?

amitsagtani97:
@jtaleric The user was exported in the environment variable but it wasn't found when fetched by the uperf. We tried to pass the user and client_ips as argument

`run_snafu --tool -user {{test_user | default("ripsaw") }} -client_ips $(hostname -I)`

Here the user was sent correctly but the client_ip wasn't sent.
and the rest of the data was saved in the ElasticSearch. Also, the CI test passed because we fetch results on elastic by index and uuid only.

`python3 ci/check_es.py -s marquez.perf.lab.eng.rdu2.redhat.com -p 9200 -u b0065619-ee93-527c-b1e1-f208e14c1f0b -i ripsaw-uperf-results`

PR #142, opened 2020-02-20
hammerd wrapper fixes for some loops

mkarg75:
Depends-On: 267


dry923:
@mkarg75 can you check the conflict with hammerdb/hammerd_wrapper.py ? Thanks!

PR #145, opened 2020-02-25
Converting datetime objects to str when dumping

aakarshg:
Have been noticing issues when trying to do a json dump or process
json documents with datetime objects present.. example would be
uperf_ts, and this help avoid the issue by defaulting conversion
to str.


jtaleric:
LGTM, good catch.

PR #146, opened 2020-02-27
Add debugging tools

rsevilla87:
We're currently using several packages from Centos8 repositories, so the best idea would be to start using Centos8 rather than ubi8 to simplify Dockerfiles and avoid using external repositories on ubi8 which might break the containers.

Container size does not differs very much from ubi8 based builds as shown in the example below:
```
$ podman images quay.io/cloud-bulldozer/uperf
REPOSITORY                      TAG      IMAGE ID       CREATED        SIZE
quay.io/cloud-bulldozer/uperf   latest   98211d2aa3fa   18 hours ago   474 MB
$ podman images quay.io/rsevilla/uperf
REPOSITORY               TAG       IMAGE ID       CREATED          SIZE
quay.io/rsevilla/uperf   centos8   82f8f7190fef   53 minutes ago   494 MB
```
In addition Centos8 ships iputils, iproute and other useful and debugging tools by default, so that would explain the extra size of the containers.

### DIFF
--- a/fio_wrapper/Dockerfile
+++ b/fio_wrapper/Dockerfile
@@ -1,9 +1,8 @@
-FROM registry.access.redhat.com/ubi8:latest
+FROM docker.io/centos:8
 
+RUN dnf install --nodocs -y python3-dnf-plugins-core
 RUN dnf copr enable ndokos/pbench -y
-COPY image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
-RUN dnf install --nodocs -y pbench-fio --enablerepo=centos8-appstream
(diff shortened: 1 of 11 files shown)


jtaleric:
instead of moving, could we offer both? 

Option 1 : Lite weight
Option 2 : Heavier, but better for debugging.

rsevilla87:
In that case we'll need to add a flag to ripsaw to run tests from debugging containers.

rsevilla87:
@jtaleric Can't see a clear use case to have two different containers. What I'm thinking now is to stop this migration to centos8 and offer more debugging tools on our current ubi8 containers. Tools like ping, ip, ss. These tools are available in the default ubi repositories, so it won't be required polluting our containers with extra package sources.

PR #149, opened 2020-03-02
Add missing es indices to tests for ycsb/pgbench

aakarshg:
This will increase code coverage for ycsb and pgbench.


dry923:
@aakarshg looks like ripsaw-pgbench-results doesn't exist in marquez

aakarshg:
@dry923 it's also odd that only pgbench test ran despite the change also impacting ycsb?

Issue #152, opened 2020-03-05
BUG: Snafu CI only triggers a single benchmark test even if PR affects more


aakarshg:
Have been noticing this oddity with snafu ci, where even if the PR changes more than a benchmark, only the first benchmark is run. 

Example PR: https://github.com/cloud-bulldozer/snafu/pull/151 this only ran fio, despite changes affecting dockerfiles of other benchmarks

dry923:
I've noticed this myself as well. Working on it atm

PR #153, opened 2020-03-05
Fixing ci tests

dry923:
This is to fix https://github.com/cloud-bulldozer/snafu/issues/152 as well as clean up a few of the ci bits.

My reasoning is similar to what we dealt with in ripsaw where the namespace was trying to be deleted multiple times and one of those failures from the wait_clean in run_ci.sh was returning an error because the namespaces was not there to be deleted. This then caused it to stop after the first test. I added in the same logic to this wait_clean as the ripsaw one where we check if it exists first. Additionally, I cleaned up some of the ordering and removed an unneeded wait_clean call.

### DIFF
--- a/ci/common.sh
+++ b/ci/common.sh
@@ -36,6 +36,7 @@ function update_operator_image() {
 }
 
 function wait_clean {
+  kubectl delete benchmark --all -n my-ripsaw
   kubectl delete all --all -n my-ripsaw
   for i in {1..30}; do
     if [ `kubectl get pods --namespace my-ripsaw | wc -l` -ge 1 ]; then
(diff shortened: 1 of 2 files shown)


aakarshg:
needs a rebase @dry923

aakarshg:
> This is to fix #152 as well as clean up a few of the ci bits.

can you please add a little bit more content to the commit/pr description explaining how this pr would help with #152 ? i don't think its obvious as i'm still trying to understand what causes #152 and how this pr will help with it?

dry923:
> > This is to fix #152 as well as clean up a few of the ci bits.
> 
> can you please add a little bit more content to the commit/pr description explaining how this pr would help with #152 ? i don't think its obvious as i'm still trying to understand what causes #152 and how this pr will help with it?

@aakarshg Updated description. Let me know if you need more. At the moment it looks like its working fine in the CI env as its on the third test or so.

PR #155, opened 2020-03-10
Install python libraries through requirements

aakarshg:
This will help push snafu to become a python library, and also
as we try to migrate other benchmarks to run_snafu and because of
imports each Dockerfile for the benchmark would be requiring more
libaries that aren't tracked in requirements.txt

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,9 @@ You must supply a "wrapper", which provides these functions:
 * build the container image for your benchmark, with all the packages, python modules, etc. that are required to run it.
 * runs the benchmark and stores the benchmark-specific results to an elasticsearch server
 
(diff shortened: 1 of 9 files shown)


dry923:
The ci system ran out of space in /. Re-running this test now.

dry923:
Looks good however needs a doc update to point out that we want to use that requirements file now.

aakarshg:
> Looks good however needs a doc update to point out that we want to use that requirements file now.

updated

rsevilla87:
@aakarshg With this change we are installing some modules twice. :(

- configparser => included by default in python3.7
- statistics => included by default in python3.7
- numpy => Installed via DNF
- requests => Installed via DNF

PR #163, opened 2020-04-01
Document size tracking

acalhounRH:
Added functionality to capture the size of each document and track the
total size. debug mode will result in logging all document sizes and
info will only capture cumulative document size.

@bengland2 
@aakarshg 
@jtaleric 
@aakarshg


acalhounRH:
@aakarshg @bengland2 Could you review this change, I did a few test runs and it works, just need review and you can merge.

acalhounRH:
@bengland2 @aakarshg @dry923 @jtaleric 

Need review/verification for this PR, this is ES + Prom integration related

bengland2:
@acalhounRH  fs-drift failed because it was only run for 5 secs, I thought we merged a fix to lengthen to 15 seconds so there would always be rates-over-time index values, which the CI looks for.   That's what failed.   Here's the PR that should have fixed this problem: https://github.com/cloud-bulldozer/ripsaw/pull/299

PR #165, opened 2020-04-13
Building uperf from src

aakarshg:
Moving away from relying on pbench rpms for uperf, given that the
rpm support will soon be going away. This would mean that we've
to pull uperf's sourcecode compile it and use it. Uperf has a
single release so we can't try to use releases, and also uperf
only has the master branch so we'll have to rely on master branch.

Depends-On: 309

### DIFF
--- a/uperf-wrapper/Dockerfile
+++ b/uperf-wrapper/Dockerfile
@@ -1,9 +1,17 @@
-FROM registry.access.redhat.com/ubi8:latest
+FROM registry.access.redhat.com/ubi8:latest as builder
 
-RUN dnf copr enable ndokos/pbench -y
-RUN dnf install -y --nodocs git python3-pip python3-numpy python3-requests python3-numpy pbench-uperf
+RUN dnf install -y --nodocs git autoconf automake gcc make
+COPY image_resources/centos8.repo /etc/yum.repos.d/centos8.repo
+RUN dnf install -y --enablerepo=centos8 lksctp-tools-devel
+RUN ln -s /usr/bin/python3 /usr/bin/python
+RUN git clone https://github.com/uperf/uperf /tmp/uperf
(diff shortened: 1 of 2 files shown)


uperf-wrapper/Dockerfile
```diff
@@ -1,7 +1,13 @@
-FROM registry.access.redhat.com/ubi8:latest
+FROM registry.access.redhat.com/ubi8:latest as builder
+
+RUN dnf install -y --nodocs git autoconf automake gcc make
+RUN ln -s /usr/bin/python3 /usr/bin/python
+RUN git clone https://github.com/uperf/uperf /tmp/uperf
+RUN pushd /tmp/uperf && autoreconf --install &&  ./configure --disable-sctp ;make
```
jtaleric, line 6: Might be interesting to include sctp due to the some of the telco asks....

aakarshg: oh yes for some reason i wasn't able to get sctp to work, because of a few dependencies ignored it for now to make sure we can move forward, will look at adding dependencies later so we can enable with sctp, unless we'd like to wait for it first.. works either way to me..

jtaleric:
`  --enable-cpc`
`  --enable-debug`

Can you please add these options?

uperf-wrapper/Dockerfile
rsevilla87: I managed to compile uperf with sctp by installing the RPM `lksctp-tools-devel`, available in the centos8 repository. The final container also needs the library `lksctp-tools` provided by the standard UBI8 repository.

aakarshg: ack i'll make changes accordingly and see if it works/

aakarshg:
closing this

PR #166, opened 2020-04-14
Bump to fio-3.19

rsevilla87:
Depends-On: 310

### DIFF
--- a/fio_wrapper/Dockerfile
+++ b/fio_wrapper/Dockerfile
@@ -1,12 +1,12 @@
 FROM registry.access.redhat.com/ubi8:latest as builder
 COPY image_resources/centos8.repo /etc/yum.repos.d/centos8.repo
 RUN dnf install -y --enablerepo=centos8 make gcc libaio zlib-devel libaio-devel
-RUN curl -L https://github.com/axboe/fio/archive/fio-3.12.tar.gz | tar xzf -
-RUN pushd fio-fio-3.12 && make -j2
+RUN curl -L https://github.com/axboe/fio/archive/fio-3.19.tar.gz | tar xzf -
+RUN pushd fio-fio-3.19 && make -j2
 
 FROM registry.access.redhat.com/ubi8:latest
-COPY --from=builder /fio-fio-3.12/fio /usr/local/bin/fio
-RUN dnf install -y --nodocs git python3-pip python3-requests python3-numpy libaio zlib procps-ng iproute net-tools ethtool nmap iputils && dnf clean all
+COPY --from=builder /fio-fio-3.19/fio /usr/local/bin/fio
+RUN dnf install -y --nodocs python3-pip python3-requests python3-numpy libaio zlib procps-ng iproute net-tools ethtool nmap iputils && dnf clean all
 RUN ln -s /usr/bin/python3 /usr/bin/python
 COPY . /opt/snafu/
 RUN pip3 install --upgrade-strategy=only-if-needed -r /opt/snafu/requirements.txt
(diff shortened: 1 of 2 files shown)


fio_wrapper/trigger_fio.py
```diff
@@ -128,7 +128,7 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
                     with open(directory + '/' + str(log_file_name), 'r') as log_file:
                         for log_line in log_file:
                             log_line_values = str(log_line).split(", ")
-                            if len(log_line_values) == 4:
+                            if len(log_line_values) == 5:
```
jtaleric, line 131: Do we want to have some backwards compatibility or simply state we support fio-3.19 (for now)... Since this won'w work with earlier versions?

rsevilla87: I guess this would be required in case of executing Snafu out from the Ripsaw context. Is that what you're thinking?

jtaleric: My thoughts exactly... We could capture the version via `fio --version` to determine the code path... Or simply document we only support X version... I am not against either.

fio_wrapper/trigger_fio.py
```diff
@@ -128,7 +128,7 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
                     with open(directory + '/' + str(log_file_name), 'r') as log_file:
                         for log_line in log_file:
                             log_line_values = str(log_line).split(", ")
-                            if len(log_line_values) == 4:
```
bengland2, line 131: what about len(log_line_values) >= 4?   Would that work for both?

fio_wrapper/trigger_fio.py
```diff
@@ -149,7 +149,8 @@ def _log_payload(self, directory, user, uuid, sample, fio_jobs_dict, fio_version
                                     (log_line_values[1]),
                                     # "nodeName": pod_details["hostname"],
                                     "data_direction": _data_direction[int(log_line_values[2])],
-                                    "offset": int(log_line_values[3])
+                                    "block_size": int(log_line_values[3]),
```
bengland2, line 152: is it critical that we parse block size?   Or is that already in the test parameters?      If not, then we could just skip it and continue to parse offset, this would work for both versions, right?

bengland2: yes it is, so we resolved to just support 1 fio version.

fio_wrapper/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
-RUN curl -L https://github.com/axboe/fio/archive/fio-3.12.tar.gz | tar xzf -
-RUN pushd fio-fio-3.12 && make -j2
+RUN curl -L https://github.com/axboe/fio/archive/fio-3.19.tar.gz | tar xzf -
+RUN pushd fio-fio-3.19 && make -j2
 
 FROM registry.access.redhat.com/ubi8:latest
-COPY --from=builder /fio-fio-3.12/fio /usr/local/bin/fio
-RUN dnf install -y --nodocs git python3-pip python3-requests python3-numpy libaio zlib procps-ng iproute net-tools ethtool nmap iputils && dnf clean all
+COPY --from=builder /fio-fio-3.19/fio /usr/local/bin/fio
+RUN dnf install -y --nodocs python3-pip python3-requests python3-numpy libaio zlib procps-ng iproute net-tools ethtool nmap iputils && dnf clean all
```
aakarshg, line 9: as @dry923 pointed out about the missing git ? ci passed fine so maybe its installed by default on ubi8?

rsevilla87: I don't know why git was there. It's not required to run the workload.

aakarshg: it was there because of the old way of cloning snafu, now we dont clone snafu anymore we just copy it

fio_wrapper/trigger_fio.py
bengland2: no it wouldn't, apparently.   resolving...

PR #169, opened 2020-04-16
Skip TLS cert verify

rsevilla87:
Add mechanism to skip TLS certificate verify step.

### DIFF
--- a/run_snafu.py
+++ b/run_snafu.py
@@ -23,6 +23,7 @@ import datetime
 import logging
 import hashlib
 import json
+import ssl
 from utils.py_es_bulk import streaming_bulk
 from utils.common_logging import setup_loggers
 from utils.wrapper_factory import wrapper_factory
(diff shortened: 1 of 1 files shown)


jtaleric:
I am alright with this change. However it isn't a *fix* for #312

rsevilla87:
> LGTM, and I'm approving because we need a fix quickly, but... would be nice if default was to NOT verify cert.

Yeah, I know. With this code, verifying cert assumes ES is exposed over a TLS endpoint. The reason behind setting it to off by default is that it would break CI since we don't have a TLS exposed ES.

aakarshg:
> Add mechanism to skip TLS certificate verify step. This PR can help to fix [cloud-bulldozer/ripsaw#312](https://github.com/cloud-bulldozer/ripsaw/issues/312)

This description will still close the issue, just a fyi.

PR #170, opened 2020-04-16
update uuid and user from list

aakarshg:
recent change led to uuid and user being treated as list for uperf, this change will fix it


mohit-sheth:
lgtm

PR #175, opened 2020-04-30
Adding Linpack

jtaleric:
Adding Linpack to SNAFU

### DIFF
--- a/linpack_wrapper/Dockerfile
+++ b/linpack_wrapper/Dockerfile
@@ -1,0 +1,10 @@
+FROM registry.access.redhat.com/ubi8:latest
+
+RUN dnf install -y --nodocs python3-pip python3-numpy python3-requests python3-numpy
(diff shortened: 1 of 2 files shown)


jtaleric:
Initial draft will not take advantage of SNAFU, simply a bash script to generate the linpack test, and kick off the harness.

linpack_wrapper/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM registry.access.redhat.com/ubi8:latest
+
+RUN dnf install -y --nodocs git python3-pip python3-numpy python3-requests python3-numpy
```
dry923, line 3: Minor nit but I don't think we need to install git w/ this unless I'm missing something.

jtaleric: fixed.

PR #176, opened 2020-05-05
Normalize UUID and user args

rsevilla87:
Depends-On: 332

### DIFF
--- a/hammerdb/hammerdb_wrapper.py
+++ b/hammerdb/hammerdb_wrapper.py
@@ -7,9 +7,6 @@ from hammerdb.trigger_hammerdb import Trigger_hammerdb
 
 class hammerdb_wrapper():
     def __init__(self, parser):
-        parser.add_argument(
-            '-u', '--uuid', nargs=1,
-            help='Provide the uuid')
         self.args = parser.parse_args()
         self.args.es_server = ""
         self.args.es_port = ""
@@ -27,6 +24,8 @@ class hammerdb_wrapper():
         self.args.db_tcp = ""
         self.args.timed_test = ""
 
+        self.args.user = os.getenv("test_user", "")
+        self.args.uuid = os.getenv("uuid", "")
         if "es_server" in os.environ:
             self.args.es_server = os.environ["es_server"]
         if "es_port" in os.environ:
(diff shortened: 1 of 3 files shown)


bengland2:
overall I appreciate the idea of refactoring to maximize common code and user experience, but I have same question that Joe had, why is uuid an env var?   It is autogenerated by ripsaw isn't it?   I think this is for the best if we standardize on how to identify a test, otherwise it becomes very difficult to write things like touchstone.

rsevilla87:
Yeah, that's the idea. UUID and user is automatically set from ripsaw but these variables might not be present in cases outside from ripsaw.

There are other variables like `clustername`, which we get from env vars.

This is just an effort to standardize the way we pass variables to snafu wrappers, so I'm open to suggestions.
 
@bengland2 can you explain with more what do you mean with "how to identify the test"?

rsevilla87:
My idea is to make a series of PR in order to add more granularity to the changes

bengland2:
are you talking about the idea of running snafu outside of ripsaw?    Because that's a very important discussion but I see no reason why test uuids couldn't be generated in the same way for that case -- the method of deploying workload generators might be different, but the way results appear in elastic search would ideally be the same.

rsevilla87:
> are you talking about the idea of running snafu outside of ripsaw? Because that's a very important discussion but I see no reason why test uuids couldn't be generated in the same way for that case -- the method of deploying workload generators might be different, but the way results appear in elastic search would ideally be the same.

The existing problem now is that some arguments are passed from the CLI itself and others from environment variables depending on the test.

bengland2:
oh, in that case why would people want to pass IN the UUID, as opposed to getting it returned to them?   If you leave it up to the user, how do you know they won't pass the same UUID in twice, defeating its uniqueness?   Don't trust the users, I am one, I should know ;-)

jtaleric:
> oh, in that case why would people want to pass IN the UUID, as opposed to getting it returned to them? If you leave it up to the user, how do you know they won't pass the same UUID in twice, defeating its uniqueness? Don't trust the users, I am one, I should know ;-)

Ah, the reason for passing "in" the UUID is to have a single UUID that is tagged from our result data to our metadata... Now, could we instead of a list of UUIDs which represent a single run, possibly.

PR #177, opened 2020-05-11
uperf cleanup

rsevilla87:
Optimize uperf code.
We still have some "print" statements at trigger_uperf.py. should we remove or change them for logger?

### DIFF
--- a/uperf_wrapper/Dockerfile
+++ b/uperf_wrapper/Dockerfile
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi8:latest
 
 RUN dnf copr enable ndokos/pbench -y
-RUN dnf install -y --nodocs git python3-pip python3-numpy python3-requests python3-numpy pbench-uperf
+RUN dnf install -y --nodocs git python3-pip pbench-uperf
 COPY image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
 RUN dnf install -y --nodocs redis --enablerepo=centos8-appstream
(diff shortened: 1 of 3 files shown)


jtaleric:
> Optimize uperf code.
> We still have some "print" statements at trigger_uperf.py. should we remove or change them for logger?

We will have to setup the logger IMHO. We don't need to know about timestamps/level/etc.

uperf_wrapper/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM registry.access.redhat.com/ubi8:latest
 
 RUN dnf copr enable ndokos/pbench -y
-RUN dnf install -y --nodocs git python3-pip python3-numpy python3-requests python3-numpy pbench-uperf
+RUN dnf install -y --nodocs git python3-pip python3-requests pbench-uperf
```
aakarshg, line 4: I guess we don't even need to install requests through rpm as it's in requirements.txt - https://github.com/cloud-bulldozer/snafu/blob/master/requirements.txt#L6

rsevilla87: We're also installing other dependencies through rpm in other tests. How do we should install them?. IMO installing them as RPM whenever is possible is safest.

aakarshg: as we're getting to close to making snafu as a package, it might make sense to move towards using pip as package manager for consistency imo.

aakarshg:
linters failing on what would be fixed by https://github.com/cloud-bulldozer/snafu/pull/178.

PR #181, opened 2020-05-21
Adds ability to calculate standard deviation for randomread and randomwrite ops

keesturam:
I believe this should fix https://github.com/cloud-bulldozer/snafu/issues/180
@acalhounRH ?

### DIFF
--- a/fio_wrapper/fio_analyzer.py
+++ b/fio_wrapper/fio_analyzer.py
@@ -134,6 +134,14 @@ class Fio_Analyzer:
                         tmp_doc['std-dev-%s' % io_size] = round(((statistics.stdev
                                                                   (average_write_result_list) /
                                                                   write_average) * 100), 3)
+                    elif 'randread' in oper:
+                        tmp_doc['std-dev-%s' % io_size] = round(
+                            ((statistics.stdev(
+                                average_read_result_list) / read_average) * 100), 3)
+                    elif 'randwrite' in oper:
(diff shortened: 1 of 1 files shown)


acalhounRH:
@bengland2 The original code does calculate the standard deviation. fio_analyzer is paired with fio_wrapper such that after all test have ran it evaluates std-dev as you have stated. 

The original code did accomplish randread and randwrite evaluation, the if statement is valid if read/write was present in the operation field (read/write/randwrite/randread). This code is just explicit checking for randread or randwrite. With this version of code the read and write checks will still be valid and the extra randread/randwrite if statements will never be invoked. 

The only limitation that this code has is that it does not process rw, 

rw, readwrite

Mixed sequential reads and writes. 

if the user is not seeing the randread or randwrite std-dev data in elasticsearch or Grafana I can work with them 1-on-1.

acalhounRH:
Recommend closing PR unless it is changed to only add sequential read/write operation.

jtaleric:
@bengland2 @acalhounRH Can we close this since it seems unnecessary (based on Alex's latest comment)?

bengland2:
@keesturam  this is getting stale, we should either cancel this or go forward, can you discuss with Alex in next couple of days if you really need this or we'll  close it?  I lost track of this one, sorry  -ben

keesturam:
Closing the PR based on Alex's inputs.

PR #173, opened 2020-04-18
Add get prometheus data to run snafu

acalhounRH:
Added code to get Prometheus data and index it into Elasticsearch. Note this PR enables the functionality but does not use it. follow-on PRs will address specific tool triggers for Prometheus data collection.

Expected data flow would be that at the conclusion of a sample loop, the tool wrapper would yield a tuple (action, "get_prometheus_trigger"), this would trigger the collection of Prometheus data. 

"action"(dict) would contain the following:

                    action: {
                              "uuid": <uuid>
                              "user": <user>
                              "clustername": <clustername>
                              "starttime": <datetime> datetime.utcnow().strftime('%s')
                              "endtime": <datetime>
                              test_config: {...}
                            }
test_config would contain the sample specific test parameters that align to the current time window.

@jtaleric @bengland2 @dry923 @aakarshg

PR #185, opened 2020-05-26
Use parallel ES indexing

rsevilla87:
In a simple 10 minutes test I got:

```
2020-05-26T15:24:29Z - INFO     - MainProcess - trigger_fio: fio has successfully finished sample 1 executing for jobname write and results are in the dir /tmp/fiod-6c4f84f7-b18d-5c45-b2f8-b251633c7612/fiojob-write-4KiB-1/1/write
2020-05-26T15:26:19Z - INFO     - MainProcess - run_snafu: Indexed results - 27466 success, 0 duplicates, 0 failures, with 0 retries.
```

and with parallel
```
2020-05-26T15:09:48Z - INFO     - MainProcess - trigger_fio: fio has successfully finished sample 1 executing for jobname write and results are in the dir /tmp/fiod-6c4f84f7-b18d-5c45-b2f8-b251633c7612/fiojob-write-4KiB-1/1/write
2020-05-26T15:10:25Z - INFO     - MainProcess - run_snafu: Indexed results - 27458 success, 0 duplicates, 0 failures, with 0 retries.
```

### DIFF
--- a/run_snafu.py
+++ b/run_snafu.py
@@ -65,6 +65,7 @@ def main():
         if os.environ["es_port"] != "":
             es['port'] = os.environ["es_port"]
             logger.info("Using elasticsearch server with port:" + es['port'])
+        parallel = int(os.getenv('parallel_indexing', 0))
     es_verify_cert = os.getenv("es_verify_cert", "true")
     if len(es.keys()) == 2:
         if os.environ["es_index"] != "":
@@ -96,7 +97,8 @@ def main():
         res_beg, res_end, res_suc, res_dup, res_fail, res_retry = streaming_bulk(es,
                                                                                  process_generator(
                                                                                      index_args,
-                                                                                     parser))
+                                                                                     parser),
+                                                                                 parallel)
 
         logger.info(
             "Indexed results - %s success, %s duplicates, %s failures, with %s retries." % (
--- a/utils/py_es_bulk.py
+++ b/utils/py_es_bulk.py
@@ -91,7 +91,7 @@ def put_template(es, name, body):
     return beg, end, retry_count
 
 
-def streaming_bulk(es, actions):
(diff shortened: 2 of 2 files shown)


bengland2:
@rsevilla87 good idea but does it scale?  If you were doing this with 30 nodes instead of 1, is py_es_bulk able to back off if ES is overloaded?   I think so (@acalhounRH what do you think?) but has anyone tried it?

aakarshg:
@portante can you please review this ? I distinctly remember you finding a boatload of problems with parallel indexing and suggested to stick with serial indexing.

portante:
Client side indexing is problematic to make scale.  Unless you control all the clients, controlling the right level of parallelism for each client can cause an Elasticsearch instance to be swamped.

If `parallel_bulk` gives the exact same semantics as `streaming_bulk`, the code will likely work.  But it is when you get into failure conditions and retries that will cause you problems.

acalhounRH:
I have already added parallel_bulk indexing with PR #173

rsevilla87:
> I have already added parallel_bulk indexing with PR #173

This one enables parallelism optionally, by default false. Do you want to wait for 173 or move forward with this one?

acalhounRH:
I would prefer to wait for it #173, if you don't mind.

aakarshg:
> I would prefer to wait for it #173, if you don't mind.

@acalhounRH can you please update your PR in that case to support enabling parallelism optionally, but defaulting to serial given @portante 's comments above.

acalhounRH:
> > I would prefer to wait for it #173, if you don't mind.
> 
> @acalhounRH can you please update your PR in that case to support enabling parallelism optionally, but defaulting to serial given @portante 's comments above.

This will require a change in both RIPSAW and snafu, ripsaw to set the env, and snafu to check the variable to switch between parallel or stream indexing.

PR #187, opened 2020-05-28
Avoid rebuilding operator and checking ES twice

rsevilla87:
Would fix https://github.com/cloud-bulldozer/snafu/issues/186 along with avoid rebuilding and pushing ripsaw in each test.

Also simplifies the function wait_clean since deleting a namespace waits for the deletion of all  objects within it by default.

Rather than doing all those "sed" invocations, we might use the image argument of the benchmark YAMLs to overwrite the image. However I'm not sure about that idea since we would have to reference each test benchmark CRD.
Thoughts?

### DIFF
--- a/ci/common.sh
+++ b/ci/common.sh
@@ -5,6 +5,7 @@ es_port=${ES_PORT:-80}
 
 default_operator_image="quay.io/benchmark-operator/benchmark-operator:master"
 
+default_ripsaw_image_prefix="quay.io/cloud-bulldozer"
 image_location=${RIPSAW_CI_IMAGE_LOCATION:-quay.io}
 image_account=${RIPSAW_CI_IMAGE_ACCOUNT:-rht_perf_ci}
 export SNAFU_IMAGE_TAG=${SNAFU_IMAGE_TAG:-snafu_ci}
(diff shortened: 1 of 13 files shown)


rsevilla87:
> IMHO We should remove the `es_check` in Ripsaw. It should be only checked via SNAFU (removing the double check).

Yeah, I get your point. However snafu's CI depends on ripsaw's CI and not the opposite. Checking ES in ripsaw's CI is the only way now to ensure changes in snafu and ripsaw don't break indexing.

jtaleric:
> > IMHO We should remove the `es_check` in Ripsaw. It should be only checked via SNAFU (removing the double check).
> 
> Yeah, I get your point. However snafu's CI depends on ripsaw's CI and not the opposite. Checking ES in ripsaw's CI is the only way now to ensure changes in snafu and ripsaw don't break indexing.

If there is a circular dependency, I wonder if we can set a var to skip indexing? by default the indexing is enabled, but if running w/ Ripsaw CI, we disable SNAFU `es_check`?

dry923:
> > > IMHO We should remove the `es_check` in Ripsaw. It should be only checked via SNAFU (removing the double check).
> > 
> > 
> > Yeah, I get your point. However snafu's CI depends on ripsaw's CI and not the opposite. Checking ES in ripsaw's CI is the only way now to ensure changes in snafu and ripsaw don't break indexing.
> 
> If there is a circular dependency, I wonder if we can set a var to skip indexing? by default the indexing is enabled, but if running w/ Ripsaw CI, we disable SNAFU `es_check`?

I think a variable for this is redundant. We are not removing the check_es script from SNAFU, we are just not calling it for the functions that are using ripsaw. If we add a new workload, foo, that is outside of ripsaw we can simply call the check_es script from foo's ci_test.sh file. Thoughts?

rsevilla87:
@jtaleric @dry923 

> > > > IMHO We should remove the `es_check` in Ripsaw. It should be only checked via SNAFU (removing the double check).
> > > 
> > > 
> > > Yeah, I get your point. However snafu's CI depends on ripsaw's CI and not the opposite. Checking ES in ripsaw's CI is the only way now to ensure changes in snafu and ripsaw don't break indexing.
> > 
> > 
> > If there is a circular dependency, I wonder if we can set a var to skip indexing? by default the indexing is enabled, but if running w/ Ripsaw CI, we disable SNAFU `es_check`?
> 
> I think a variable for this is redundant. We are not removing the check_es script from SNAFU, we are just not calling it for the functions that are using ripsaw. If we add a new workload, foo, that is outside of ripsaw we can simply call the check_es script from foo's ci_test.sh file. Thoughts?

+1 This PR simplifies the overall CI besides of preventing double-checking of ES indices.

dry923:
@rsevilla87 CI from the last run is showing an error when building the benchmark-operator:snafu_ci

```
00:00:02.446 + image_spec=quay.io/rht_perf_ci/benchmark-operator:snafu_ci
00:00:02.446 + operator-sdk build quay.io/rht_perf_ci/benchmark-operator:snafu_ci --image-builder podman
00:00:02.511 time="2020-07-17T21:15:49Z" level=info msg="Building OCI image quay.io/rht_perf_ci/benchmark-operator:snafu_ci"
00:00:02.704 STEP 1: FROM quay.io/operator-framework/ansible-operator:v0.17.1
00:00:02.771 STEP 2: USER root
00:00:02.775 --> Using cache 38af728f984f1717973bd2c0bb81eea46d753c57947ed6d371ee4c2edec81132
00:00:02.796 STEP 3: COPY requirements.yml ${HOME}/requirements.yml
00:00:04.750 --> d570bdbb4787bdd21961591e4194cdbbe08405c3142e6877b94b8c8a6e082169
00:00:04.770 STEP 4: RUN ansible-galaxy collection install -r ${HOME}/requirements.yml  && chmod -R ug+rwx ${HOME}/.ansible
00:00:06.348 Process install dependency map
00:00:06.348 |/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-\|/-ERROR! Unknown error when attempting to call Galaxy at 'https://galaxy.ansible.com/api/': <urlopen error [Errno -2] Name or service not known>
00:00:16.105 Error: error building at STEP "RUN ansible-galaxy collection install -r ${HOME}/requirements.yml  && chmod -R ug+rwx ${HOME}/.ansible": error while running runtime: exit status 1
00:00:16.111 Error: failed to output build image quay.io/rht_perf_ci/benchmark-operator:snafu_ci: (failed to exec []string{"podman", "build", "-f", "build/Dockerfile", "-t", "quay.io/rht_perf_ci/benchmark-operator:snafu_ci", "."}: exit status 125)
00:00:16.111 Usage:
00:00:16.111   operator-sdk build <image> [flags]
00:00:16.111 
00:00:16.111 Flags:
00:00:16.111       --go-build-args string      Extra Go build arguments as one string such as "-ldflags -X=main.xyz=abc"
00:00:16.111   -h, --help                      help for build
00:00:16.111       --image-build-args string   Extra image build arguments as one string such as "--build-arg https_proxy=$https_proxy"
00:00:16.111       --image-builder string      Tool to build OCI images. One of: [docker, podman, buildah] (default "docker")
00:00:16.111 
00:00:16.111 Global Flags:
00:00:16.111       --verbose   Enable verbose logging
00:00:16.111 
```

ci/common.sh
```diff
@@ -5,6 +5,7 @@ es_port=${ES_PORT:-80}
 
 default_operator_image="quay.io/benchmark-operator/benchmark-operator:master"
 
+default_ripsaw_image_spec="quay.io/cloud-bulldozer"
```
jtaleric, line 8: just a small nit, this is not a spec, but an org.

rsevilla87: Replaced by default_ripsaw_image_prefix

PR #193, opened 2020-07-08
Add simple method to pass vegeta result json

jtaleric:
Allow the user to run vegeta outside of SNAFU, but allow us to pass the Vegeta json result to SNAFU for processing.

### DIFF
--- a/vegeta_wrapper/trigger_vegeta.py
+++ b/vegeta_wrapper/trigger_vegeta.py
@@ -31,8 +31,18 @@ class Trigger_vegeta():
         self.duration = args.duration
         self.cluster_name = args.cluster_name
         self.keepalive = args.keepalive
+        if args.results :
+            self.target_name = args.target_name
+            self.results = args.results
+        else :
+            self.results = None
 
     def _json_payload(self, data, sample):
+        if self.results:
+            targets = self.target_name
+        else :
+            targets = os.path.basename(self.targets)
+
         payload = {
             "workload": "vegeta",
             "uuid": self.uuid,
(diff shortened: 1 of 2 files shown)


jtaleric:
https://search-cloud-perf-lqrf3jjtaqo7727m7ynd2xyt4y.us-west-2.es.amazonaws.com/_plugin/kibana/app/kibana#/doc/31536180-abc0-11ea-8f23-015cc92edd9b/ripsaw-vegeta-results/_doc?id=7445d832be3442944566fb3be26adc94&_g=(refreshInterval:(pause:!t,value:0),time:(from:now-15h,to:now))

vegeta_wrapper/trigger_vegeta.py
```diff
@@ -31,20 +31,39 @@ def __init__(self, args):
...
+                "uuid": self.uuid,
+                "user": self.user,
+                "cluster_name": self.cluster_name,
+                "iteration": sample,
+                "duration": self.duration,
+                "workers": self.workers,
+                "keepalive": self.keepalive,
+                "targets": os.path.basename(self.targets),
+                "hostname": socket.gethostname()
+            }
```
rsevilla87, line 66: This could be simplified with something like:

```suggestion
        if self.results:
            targets = serf.target_name 
        else:
            targets = os.path.basename(self.targets)
        payload = {
            "workload": "vegeta",
            "uuid": self.uuid,
            "user": self.user,
            "cluster_name": self.cluster_name,
            "iteration": sample,
            "duration": self.duration,
            "workers": self.workers,
            "keepalive": self.keepalive,
            "targets": targets,
            "hostname": socket.gethostname()
        }
```

jtaleric: :+1: Thanks!

PR #196, opened 2020-07-17
Drop cache

bengland2:
This change would enable snafu benchmarks to utilize privileged cache-dropper pods as a cherrypy web service with well-known TCP ports.   run_snafu.py would invoke drop_cache() before every sample.  The ripsaw component of the benchmark would start cache dropper pods and then would wait until they were running, then get their IP addresses and pass them as environment variables into run_snafu.py.   It was tested using minikube - run_snafu has to be run within a k8s pod to be able to talk to cache-dropper pod.


run_snafu.py
```diff
@@ -128,9 +129,11 @@ def main():
 def process_generator(index_args, parser):
     benchmark_wrapper_object_generator = generate_wrapper_object(index_args, parser)
 
+    drop_cache()
     for wrapper_object in benchmark_wrapper_object_generator:
         for data_object in wrapper_object.run():
             for action, index in data_object.emit_actions():
+                drop_cache()  # before the next
```
rsevilla87, line 136: This will drop cache at the beginning of snafu. Shouldn't be dropped for each sample. The same as ceph cache dropping: https://github.com/cloud-bulldozer/snafu/blob/master/fio_wrapper/fio_wrapper.py#L73-L74

bengland2: @rsevilla87 thanks for taking a look.   I disagree - emit_actions() is what actually runs a sample and yields objects for elasticsearch.   So by placing drop_cache() after the emit_actions call we are ensuring that cache is dropped before the *next* sample, and we call drop_cache() before the first sample is collected as well.  What I am concerned about is that emit_actions() yields > 1 document per sample, so this could result in excessive cache dropping.   I couldn't see a good way to get drop_cache() to run exactly once per sample.   @acalhounRH ?

bengland2:
[evidence that it works](https://pastebin.com/ydyEf6Je)

src/utils/request_cache_drop.py
```diff
@@ -0,0 +1,64 @@
...
+    # drop Ceph OSD cache if requested to
+    
+    ceph_cache_drop_pod_ip = os.getenv('ceph_drop_pod_ip')
+    logger.info('ceph OSD cache drop pod: %s' % str(ceph_cache_drop_pod_ip))
+    if ceph_cache_drop_pod_ip != None:
+        conn = http.client.HTTPConnection(ceph_cache_drop_pod_ip, port=dropKernelCachePort, timeout=http_timeout)
+        if http_debug_level > 0:
+            conn.set_debuglevel(http_debug_level)
+        logger.info('requesting ceph to drop cache via %s:%d' % (ip, dropCephCachePort))
+        conn.request('GET', '/DropCephCache')
```
rsevilla87, line 40: Where is this endpoint implemented?

src/utils/request_cache_drop.py
```diff
@@ -0,0 +1,64 @@
...
+
+http_debug_level = int(os.getenv('HTTP_DEBUG_LEVEL', default=0))
+
+http_timeout = 10
+
+def drop_cache():
+
+    # drop Ceph OSD cache if requested to
+    
+    ceph_cache_drop_pod_ip = os.getenv('ceph_drop_pod_ip')
```
rsevilla87, line 33: This env var should be `ceph_cache_drop_pod_ip` as stated in https://github.com/cloud-bulldozer/snafu/blob/master/src/fio_wrapper/fio_wrapper.py#L55

rsevilla87: Now that ceph cache drop is implemented in request_cache_drop.py I think we can get rid of the old code:
https://github.com/cloud-bulldozer/snafu/blob/master/src/fio_wrapper/fio_wrapper.py#L71-L81

What do you think?

bengland2: correct, was going to do that in the next PR that was going to do the same thing to ceph cache dropping - pull it out of individual benchmark and make it common code usable by any benchmark that wants it.

bengland2: Is there any way to pull out the steps in roles/fio_distributed/tasks/main.yml that start the kernel cache drop pods and capture the kernel cache drop pod IPs and make them a "subroutine" that gets called by the benchmark instead?  Would that be a role in ansible?

rsevilla87:
Hey @bengland2, can you rebase please?

bengland2:
@rsevilla87 I rebased, what do you think?

bengland2:
am cleaning up the lint errors now, will rerun afterwards.    My bad, forgot to run pycodestyle

bengland2:
actually the linter found bugs, so worth the effort, retesting with changes, then will re-CI it, then will merge.

bengland2:
I'm not buying that this particular PR broke only sysbench and nothing else.   Merging it.    It does run fio.

jtaleric:
```22:21:22 ++ echo 'Logs from sysbench-1ecb6eab-sc4xt'
22:21:22 Logs from sysbench-1ecb6eab-sc4xt
22:21:22 ++ kubectl -n my-ripsaw-6c45a80f-3f6b-446e-8cc4-48a4e978585b logs --tail=40 sysbench-1ecb6eab-sc4xt
22:21:22 Error from server (BadRequest): container "sysbench" in pod "sysbench-1ecb6eab-sc4xt" is waiting to start: PodInitializing
22:21:22 +++ error```

I am curious, most of our other CI has retries, but it looks like SNAFU does not?

PR #199, opened 2020-07-30
added mysql and postgresql functionality

mkarg75:
I've added initial support for mysql and postgres DB support. This PR needs to be merged before I can file the PR for ripsaw.


src/image_resources/centos8-appstream.repo
```diff
@@ -1,5 +1,5 @@
 [centos8-appstream]
 name=CentOS-8-Appstream
 baseurl=http://mirror.centos.org/centos/8/AppStream/x86_64/os/
-enabled=0
```
rsevilla87, line 4: Why did you enable centos repos by default?. We left them disabled to prevent RPM conflicts.
You can enable them with dnf passing either --enablerepo=centos8 or --enablerepo=centos8-appstream

mkarg75:
Creating a new one instead of this.

PR #200, opened 2020-08-06
Disable native CPU detection

rsevilla87:
Disable native CPU detection from FIO building to support CPUs with different instruction sets.

[Fedora RPMs are built ](https://kojipkgs.fedoraproject.org//packages/fio/3.19/3.fc32/data/logs/x86_64/build.log) with the flag `--disable-optimizations` which essentially gives the same result.


ekuric:
@jtaleric as @bengland2 can you review this change and ack/nack so it is merged to get new image.

PR #211, opened 2020-09-16
Add determining if scale up/down and gather initial worker count

dry923:
The action will now be defined as:

- scale_up
- scale_down
- scale_nochange

We also gather the initial worker count and index the field as: init_worker_count


mohit-sheth:
maybe index clusterversion too?
have pre and post cerberus hook like we have for /workloads?

dry923:
> maybe index clusterversion too?

I skipped the cluster version since its in other metrics and not currently in what @jtaleric with the other cluster timings.

> have pre and post cerberus hook like we have for /workloads?
I'm adding that into the e2e testing. Felt redundant to add it in here as well.

jtaleric (approved):
LGTM

rsevilla87 (approved):
LGTM

PR #212, opened 2020-09-16
Update CI test scripts

dry923:
Updated CI scripts for:

- Remove unused check_es.py script and function
- Update UUID generation to shorten the length of the string in case we have long test names
- Moved updating the ci/common.sh file. This was being updated before the git diff so it was always being flagged to run a full test. Moving it to after should resolve it.


jtaleric (approved):
LGTM

rsevilla87 (approved):

PR #221, opened 2020-10-22
Install dumb-init through rpm

smalleni:


snafu/cyclictest_wrapper/Dockerfile
```diff
@@ -5,7 +5,7 @@ RUN dnf install -y --nodocs python3 python3-pip python3-devel numactl-devel perl
 RUN dnf -y install https://www.rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/rt-tests-1.8-11.el8.x86_64.rpm
 RUN dnf -y --enablerepo=extras install --nodocs wget tmux \
     && dnf install -y --nodocs stress-ng \
```
jtaleric, line 7: Can we break these out?  instead of adding `&&`

smalleni: Done

PR #231, opened 2020-11-18
need to add realpath to resolve relative path to util scripts.

acalhounRH:
with only using dirname on __file__ and __file__ only being a script name the
directory path resulted in no resolution to the relative path, we can
resolve this by adding realpath.

@dry923 
@rsevilla87 
@jtaleric


rsevilla87 (approved):
LGTM

dry923 (approved):
LGTM!

PR #237, opened 2020-12-09
Adding support for clusters (like ROSA) that are labeled differently

dry923:
ROSA created clusters are labeled differently. This adds support for other types of labelieng.

ROSA Cluster labels:

```
labels:
  hive.openshift.io/machine-pool: worker
  hive.openshift.io/managed: "true"

```

While the script currently expects:

```
labels:
  node-role.kubernetes.io/worker: ""
```


mohit-sheth (approved):
LGTM

chaitanyaenr (approved):
LGTM

PR #239, opened 2020-12-14
[Scale Job] Changed worker machineset query to use name 

dry923:
The worker query while waiting for machinesets to have all their nodes was previously looking at what is NOT an infra or workload node. This is flawed as sometimes a worker can have multiple roles. This change uses the worker machine set name instead of a NOT list. This also allows managed services clusters better functionality.


mohit-sheth (approved):
LGTM

rsevilla87 (approved):
LGTM

PR #241, opened 2021-01-07
run_id should be optional

jtaleric:
Making run_id optional

### DIFF
--- a/snafu/run_snafu.py
+++ b/snafu/run_snafu.py
@@ -49,7 +49,8 @@ def main():
     parser.add_argument(
         '-t', '--tool', help='Provide tool name', required=True)
     parser.add_argument(
-        '--run-id', help='Run ID to unify benchmark results in ES'
+        '--run-id', help='Run ID to unify benchmark results in ES',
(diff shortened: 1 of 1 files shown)


mohit-sheth (approved):
LGTM

dry923:
If we're defaulting it then line 181 will always be true and the if portion can be removed.

```
    if index_args.run_id:
        logger.debug(f"Run ID is {index_args.run_id}")
        es_valid_document['run_id'] = action['run_id'] = index_args.run_id
```

jtaleric:
> If we're defaulting it then line 181 will always be true and the if portion can be removed.
> 
> ```
>     if index_args.run_id:
>         logger.debug(f"Run ID is {index_args.run_id}")
>         es_valid_document['run_id'] = action['run_id'] = index_args.run_id
> ```
> 
> +1 here, otherwise lgtm

yeah, I default the string to NA. since we will always have the key/value pair in the ES doc.

rsevilla87 (approved):
LGTM!

PR #245, opened 2021-01-12
Simple fix to handle case where there is no es server specified

smalleni:

### DIFF
--- a/snafu/run_snafu.py
+++ b/snafu/run_snafu.py
@@ -68,7 +68,7 @@ def main():
     es_settings = {}
     es_settings["server"] = os.getenv("es")
     es_settings["verify_cert"] = os.getenv("es_verify_cert", "true")
-    if ":443" in es_settings["server"] :
+    if es_settings["server"] and ":443" in es_settings["server"]:
         es_settings["verify_cert"] = "false"
     if es_settings["server"]:
         index_args.prefix = os.getenv("es_index", "")


jtaleric (approved):
LGTM, thanks Sai.

snafu/run_snafu.py
```diff
@@ -68,7 +68,7 @@ def main():
     es_settings = {}
     es_settings["server"] = os.getenv("es")
     es_settings["verify_cert"] = os.getenv("es_verify_cert", "true")
-    if ":443" in es_settings["server"] :
+    if es_settings["server"] and ":443" in es_settings["server"]:
```
rsevilla87, line 71: Why this change?? the ES logic only gets executed if "server" env var is specified:
https://github.com/cloud-bulldozer/benchmark-wrapper/blob/ce249862486c78f806a93995a4bc4a612169861c/snafu/run_snafu.py#L73-L80

jtaleric, line 71: hm, should this be `if "server" in es_settings` ?

rsevilla87: os.getenv("es") returns None if "es" is not set

smalleni: Yes but you will see `TypeError: argument of type 'NoneType' is not iterable` when trying to check if ":443" in...

PR #249, opened 2021-02-08
TRex wrapper script for testpmd workload

mukrishn:
This is required for - TestPMD workload https://github.com/cloud-bulldozer/benchmark-operator/pull/510


snafu/trex_wrapper/ci_test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
smalleni, line 1: Why even have this when we are not testing in CI?
@mukrishn

mukrishn: I will remove this file, i  was not sure at that time.

smalleni (approved):

snafu/trex_wrapper/scripts/run_simple_burst
```diff
@@ -0,0 +1,221 @@
...
+                      "tx_pps": stats["global"]["tx_pps"],
+                      "rx_bps": stats["global"]["rx_bps"],
+                      "tx_bps": stats["global"]["tx_bps"],
+                      "rx_drop_bps": stats["global"]["rx_drop_bps"],
+                      "cpu_util": stats["global"]["cpu_util"],
+                      "ts_epoch": current_ts + i
+                      })
+
+        # wait_on_traffic was not good for 2MPPS or higher, instead doing reset
+        # self.client.wait_on_traffic(ports=self.ports)
```
jtaleric, line 155: remove this since it is commented out?

snafu/trex_wrapper/scripts/run_simple_burst
```diff
@@ -0,0 +1,221 @@
...
+        log.info("destination macs available")
+        for idx in range(0, self.num_ports):
+            self.add_streams(idx, macs[idx])
+
+    def collect_stats(self, duration):
+        output = []
+        current_ts = time.time()
+
+        for i in range(0,duration):
+            time.sleep(1)
```
jtaleric, line 137: lets allow the user to provide a collection interval... and add this to the metadata. Long running zero drop tests could generate a lot of data... if we need that sort of granularity, that is fine, but there might be a use to allow a 10 second interval.

mukrishn: Yes, it make sense to make this configurable based on the test load.

jtaleric (approved):
LGTM

PR #250, opened 2021-02-08
Uperf scale variables

mrbojangles3:
These changes add the new variables from the uperf-scale enhancements to benchmark-operator. Since there are defaults provided in the python calls, I assume this change is not dependent on the forthcoming pr to benchmark operator. I am also planning on squash merging, as the intermediate commit messages are not always descriptive.

### DIFF
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -82,8 +82,8 @@ class Trigger_uperf():
                 "nodes_in_iter":int(self.nodes_in_iter),
                 "step_size" : self.step_size,
                 "colocate" : self.colocate,
-                "density_range" : self.density_range,
-                "node_range" : self.node_range,
+                "density_range" : [int(x) for x in self.density_range.split('-')],
+                "node_range" : [int(x) for x in self.node_range.split('-')],
                 "pod_id" : int(self.pod_id)
 
             }
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -37,6 +37,9 @@ class Trigger_uperf():
         self.num_pairs = args.num_pairs
         self.multus_client = args.multus_client
         self.networkpolicy = args.networkpolicy
+        self.nodes_in_iter = args.nodes_in_iter
+        self.pod_density = args.pod_density
+        self.node_id = args.node_id
 
     def _json_payload(self, results, data, sample):
         processed = []
@@ -97,8 +100,8 @@ class Trigger_uperf():
         # We assume message_size=write_message_size to prevent breaking dependant implementations
         return results, {"test_type": test_type, "protocol": protocol, "message_size": int(wsize),
                          "read_message_size": int(rsize), "num_threads": int(nthr), "duration": len(results),
-                         "density": int(self.args.pod_density), "nodes_in_iter":int(self.args.nodes_in_iter),
-                         "node_name": self.args.node_id}
+                         "density": int(self.pod_density), "nodes_in_iter":int(self.nodes_in_iter),
+                         "node_name": self.node_id}
 
     def emit_actions(self):
         if not os.path.exists(self.workload):
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -41,6 +41,8 @@ class Trigger_uperf():
         self.pod_density = args.pod_density
         self.colocate = args.colocate
         self.step_size = args.step_size
+        self.density_range = args.density_range
+        self.node_range = args.node_range
 
     def _json_payload(self, results, data, sample):
         processed = []
@@ -79,6 +81,8 @@ class Trigger_uperf():
                 "nodes_in_iter":int(self.nodes_in_iter),
                 "step_size" : self.step_size,
                 "colocate" : self.colocate,
+                "density_range" : self.density_range,
+                "node_range" : self.node_range,
 
             }
             datapoint.update(data)
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -43,6 +43,7 @@ class Trigger_uperf():
         self.step_size = args.step_size
         self.density_range = args.density_range
         self.node_range = args.node_range
+        self.pod_id = args.pod_id
 
     def _json_payload(self, results, data, sample):
         processed = []
@@ -83,6 +84,7 @@ class Trigger_uperf():
                 "colocate" : self.colocate,
                 "density_range" : self.density_range,
                 "node_range" : self.node_range,
+                "pod_id" : int(self.pod_id)
 
             }
             datapoint.update(data)
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -55,7 +55,7 @@ class Trigger_uperf():
                 "user": self.user,
                 "cluster_name": self.cluster_name,
                 "hostnetwork": self.hostnetwork,
-                "iteration": sample,
+                "teration": sample,
                 "remote_ip": self.remoteip,
                 "client_ips": self.clientips,
                 "uperf_ts": datetime.fromtimestamp(int(result[0].split('.')[0]) / 1000),
@@ -96,7 +96,9 @@ class Trigger_uperf():
         results = re.findall(r"timestamp_ms:(.*) name:Txn2 nr_bytes:(.*) nr_ops:(.*)", stdout)
         # We assume message_size=write_message_size to prevent breaking dependant implementations
         return results, {"test_type": test_type, "protocol": protocol, "message_size": int(wsize),
-                         "read_message_size": int(rsize), "num_threads": int(nthr), "duration": len(results)}
+                         "read_message_size": int(rsize), "num_threads": int(nthr), "duration": len(results),
+                         "density": int(self.args.pod_density), "nodes_in_iter":int(self.args.nodes_in_iter),
+                         "node_name": self.args.node_id}
 
     def emit_actions(self):
         if not os.path.exists(self.workload):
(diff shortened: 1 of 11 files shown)


snafu/uperf_wrapper/trigger_uperf.py
```diff
@@ -37,6 +37,13 @@ def __init__(self, args):
         self.num_pairs = args.num_pairs
         self.multus_client = args.multus_client
         self.networkpolicy = args.networkpolicy
+        self.nodes_in_iter = args.nodes_in_iter
```
smalleni, line 40: I'd prefer to make this variable something more descriptive as nodes_in_iteration.

snafu/uperf_wrapper/uperf_wrapper.py
```diff
@@ -54,6 +54,13 @@ def __init__(self, parent_parser):
         self.args.num_pairs = os.getenv("num_pairs", "")
         self.args.multus_client = os.getenv("multus_client", "")
         self.args.networkpolicy = os.getenv("networkpolicy", "")
+        self.args.nodes_in_iter = os.getenv("node_count", "0")
```
smalleni, line 57: I think we should just set it to empty as default, any reason or 0s? Does ES complain if its an empty field?

snafu/uperf_wrapper/uperf_wrapper.py
```diff
@@ -54,6 +54,13 @@ def __init__(self, parent_parser):
...
+        self.args.pod_density = os.getenv("pod_count", "0")
+        self.args.colocate = os.getenv("colocate","False")
+        self.args.step_size = os.getenv("stepsize","")
+        self.args.density_range = os.getenv("density_range","0-0")
+        self.args.node_range = os.getenv("node_range","0-0")
+        self.args.pod_id = os.getenv("my_pod_idx","0")
```
smalleni, line 63: Leave empty if not defined.

snafu/uperf_wrapper/uperf_wrapper.py
```diff
@@ -54,6 +54,13 @@ def __init__(self, parent_parser):
         self.args.num_pairs = os.getenv("num_pairs", "")
         self.args.multus_client = os.getenv("multus_client", "")
         self.args.networkpolicy = os.getenv("networkpolicy", "")
+        self.args.nodes_in_iter = os.getenv("node_count", "0")
+        self.args.pod_density = os.getenv("pod_count", "0")
+        self.args.colocate = os.getenv("colocate","False")
+        self.args.step_size = os.getenv("stepsize","")
+        self.args.density_range = os.getenv("density_range","0-0")
+        self.args.node_range = os.getenv("node_range","0-0")
```
smalleni, line 62: Aren't we exporting a list like [2,4]? if so, lets directly set node_range to a list type instead of string.

snafu/uperf_wrapper/uperf_wrapper.py
```diff
@@ -54,6 +54,13 @@ def __init__(self, parent_parser):
         self.args.num_pairs = os.getenv("num_pairs", "")
         self.args.multus_client = os.getenv("multus_client", "")
         self.args.networkpolicy = os.getenv("networkpolicy", "")
+        self.args.nodes_in_iter = os.getenv("node_count", "0")
+        self.args.pod_density = os.getenv("pod_count", "0")
+        self.args.colocate = os.getenv("colocate","False")
+        self.args.step_size = os.getenv("stepsize","")
+        self.args.density_range = os.getenv("density_range","0-0")
```
jtaleric, line 61: Can you add a comment on what this 0-0 means?

snafu/uperf_wrapper/uperf_wrapper.py
jtaleric, line 62: Same here

snafu/uperf_wrapper/uperf_wrapper.py
jtaleric, line 63: What is my_pod_idx?

jtaleric:
Also, can we squash?

mrbojangles3:
Usually you can squash merge, its a dropdown arrow inside the green commit button.

snafu/uperf_wrapper/trigger_uperf.py
mrbojangles3: I thought `iter` was a sufficient and well known abbreviation, its already a long variable.

snafu/uperf_wrapper/uperf_wrapper.py
mrbojangles3: I was going for a sentinel value, that we would know if it wasn't set in the config file. Empty string works as well, but I figured ES returning results with a value that is known as an error code would be useful.  Will change.

snafu/uperf_wrapper/uperf_wrapper.py
smalleni: @mrbojangles3 , this is a list kind of field from the CR
node_range: [2,4] for example. I'm of the opinion we keep this as a list/array/ here as well, not sure about the "0-0" kind of string notation.

smalleni: Also, to restate if we dont have the field default should be empty.

snafu/uperf_wrapper/trigger_uperf.py
smalleni: Yeh, I'm not sure about the well known abbreviation part, maybe we can ask someone else to weight in. Not married to any name.

snafu/uperf_wrapper/uperf_wrapper.py
mrbojangles3: I was planning on using this as metadata for a run, not planning on using the range feature of ES.

smalleni: I don't see how having this in the form you have current will help with anything in the future. even to do run tor un compairsons with touchstone or the likes in future, I'd prefer to keep node_range and desnity_range as lists.

jtaleric:
> Usually you can squash merge, its a dropdown arrow inside the green commit button.

yes! I have seen that glorious button! I actually have used it in the past too! for a simple change that doesn't seem to add a considerable amount of code, I thought a squash was simple...

snafu/uperf_wrapper/uperf_wrapper.py
mrbojangles3: @smalleni Can you elaborate why? I ask because they are exported as bash environment variables which doesn't have a list type that I am aware of so the list aspect is gone. I don't know of a way that the list aspect would be useful in elastic search. After having gone though some iterations with benchmark-wrapper and benchmark-comparison the uuid will serve the purpose of identifying a run so maybe we don't need the variables ?

mrbojangles3:
@jtaleric maybe I should rebase after making changes suggested here, hopefully the magic button will appear.

snafu/uperf_wrapper/uperf_wrapper.py
smalleni: As long as they are exported you can always slice and dice convert them to a list in python :-) 

I’m thinking of cases where you want to identify runs that had a node_range of a particular range. We don’t always search by UUID, we also search by these field sometimes to get all runs having there search Params, if that makes sense. I’m fine if you don’t think we need it but I just like to preserve data in useful formats. Saving a range as a string didn’t make sense to me and I don’t agree with your argument that since this is an exported bash variable you can’t convert it to a list using string manipulations :)

mrbojangles3: Do we have something that works like you are describing today? I share the desire to keep data in useful formats, but I am not sure of the use case you are describing. If we keep this as a string we can still look for the strings that match for example 20-25 for node density.

smalleni: Oh, I can show you plenty of examples f sending a list/array over to ES, just maybe not in this repo. Check out the raw field in the docs at http://ospperf8.mobius.lab.eng.rdu2.redhat.com:5601/goto/0588c9b3b23122473acdbe3a9b790263 which are an array type. My whole point is
1. Just because something is exported as a bash env variable, doesn't mean we can parse it to typecaste to a list
2. I would like to see integrity of data types preserved
3. There is no such special range/array type in ES. Any field accepts a list of values as long as they are of the same type https://www.elastic.co/guide/en/elasticsearch/reference/current/array.html
4. The biggest and most important reason to have this as a list is, we need to start thinking about usecases other than where you already know the UUID and are searching results based on that. Think of the inverse, you have a bunch of data in ES and want to get the UUIDs of all runs that had a certain node range etc. While you can match on a string as well, having a list of integers gives you the flexibility to query and filter things based on range as well : https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html

smalleni (changes requested):
Replied on the comment about why I would like to see the ranges as lists/arrays.

smalleni (approved):
Thank you for making the requested changes, LGTM

smalleni (approved):

snafu/uperf_wrapper/trigger_uperf.py
```diff
@@ -70,7 +77,15 @@ def _json_payload(self, results, data, sample):
                 "server_node": self.server_node,
                 "num_pairs": self.num_pairs,
                 "multus_client": self.multus_client,
-                "networkpolicy": self.networkpolicy
+                "networkpolicy": self.networkpolicy,
+                "density": int(self.pod_density),
+                "nodes_in_iter":int(self.nodes_in_iter),
```
chinthagovardhan, line 82: this gives error as pod_density default value is ""

snafu/uperf_wrapper/uperf_wrapper.py
```diff
@@ -54,6 +54,18 @@ def __init__(self, parent_parser):
         self.args.num_pairs = os.getenv("num_pairs", "")
         self.args.multus_client = os.getenv("multus_client", "")
         self.args.networkpolicy = os.getenv("networkpolicy", "")
+        self.args.nodes_in_iter = os.getenv("node_count", "")
+        self.args.pod_density = os.getenv("pod_count", "")
```
chinthagovardhan, line 58: initialize pod_count to default value 0 to avoid above type conversion error

PR #251, opened 2021-02-16
Handle case where some values don't exist gracefully

smalleni:

### DIFF
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -78,14 +78,13 @@ class Trigger_uperf():
                 "num_pairs": self.num_pairs,
                 "multus_client": self.multus_client,
                 "networkpolicy": self.networkpolicy,
-                "density": int(self.pod_density or 0),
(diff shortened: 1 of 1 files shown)


jtaleric:
Only nit is the title, benchmark-wrapper doesn't deal in CRs... but I understand what you are saying ;)

smalleni:
> Only nit is the title, benchmark-wrapper doesn't deal in CRs... but I understand what you are saying ;)

UGH :-)

HughNhan:
The patch fixed the issue on my local test.

rsevilla87:
Given the merget patch at https://github.com/cloud-bulldozer/benchmark-wrapper/pull/252 , I think we don't need this anymore

smalleni:
> Given the merged patch at #252 , I think we don't need this anymore

Yep, I guess a new PR was opened without merging this ;) @rsevilla87 I do have some questions though, for fields where values do not exist why are we assuming it to be 0? I think that is misleading, can we just not send "" ?

mohit-sheth (approved):
LGTM

jtaleric (approved):
lgtm

PR #254, opened 2021-02-18
Adding log generator script and container file

dry923:
Depends-On: 545

# Description
Adding a log generator workload that will generate X messages of Y size (in bytes) per second/minute over a duration of Z minutes. If provided it will also check the backed log aggregator to confirm it received the expected number of messages. Supported backends are elasticsearch and AWS cloudwatch.

NOTE: This requires a backend storage to be deployed if we want to create a valid ci test for this.

benchmark-operator PR: https://github.com/cloud-bulldozer/benchmark-operator/pull/545

### DIFF
--- a/snafu/log_generator_wrapper/trigger_log_generator.py
+++ b/snafu/log_generator_wrapper/trigger_log_generator.py
@@ -18,7 +18,6 @@ from time import sleep
 import random
 import string
 import logging
-import redis
 import boto3
 import json
 import subprocess
@@ -36,10 +35,7 @@ class Trigger_log_generator():
         self.messages_per_second = args.messages_per_second
         self.duration = args.duration
         self.pod_count = args.pod_count
-        self.redisip = args.redisip
-        self.redisport = args.redisport
         self.pod_name = args.pod_name
-        self.namespace = args.namespace
         self.timeout = args.timeout
         self.cloudwatch_log_group = args.cloudwatch_log_group
         self.aws_access_key = args.aws_access_key
@@ -118,17 +114,6 @@ class Trigger_log_generator():
                 count += self.messages_per_second
         return count
 
-    def _wait_for_pods(self,my_redis,channel):
-        subchannel = my_redis.pubsub()
-        subchannel.subscribe(channel)
-        subscribers = dict(my_redis.pubsub_numsub(channel))[channel]
-        while subscribers < self.pod_count:
-            sleep(1)
-            subscribers = dict(my_redis.pubsub_numsub(channel))[channel]
-        # Sleep 10 to ensure all pods are ready and allow any lingering messages to clear
-        sleep(10)
-        subchannel.unsubscribe(channel)
-
     def _check_cloudwatch(self,start_time,end_time):
         logger.info("Checking CloudWatch for expected messages")
         if self.aws_access_key is not None and self.aws_secret_key is not None:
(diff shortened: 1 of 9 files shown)


jtaleric:
I like the idea of having this part of snafu/benchmark-wrapper, here are some more ideas for this work - 

1. Add method to "check", the expected number of logs have landed in the expected back-end. Today, it is Elastic, tomorrow CloudWatch?
2. Index results from the execution of the workload  -- How rate, message size, back-end, and messages received 
3. This can easily be scaled with benchmark-operator. We might want to consider the 1:N `log-generator server` and `log-generator workers`.

dry923:
@chaitanyaenr @jtaleric @rsevilla87 out of draft and good to review please.

snafu/log_generator_wrapper/trigger_log_generator.py
```diff
@@ -0,0 +1,274 @@
...
+            my_redis = redis.StrictRedis(
+                host=self.redisip,
+                port=self.redisport,
+                charset="utf-8",
+                decode_responses=True)
+            channel = "log-generator-" + self.uuid
+            self._wait_for_pods(my_redis,channel)
+
+        timestamp = time.strftime("%Y-%m-%dT%H:%M:%S")
+        start_time = time.time()
```
rsevilla87, line 225: Maybe it's better to use datetime.datetime.now() ?, I'm suggesting this because this way you can subtract dates natively (w/o converting them to epoch)

```python
start_time = datetime.datetime.now()
message_count = self._run_log_test()
end_time = datetime.datetime.now()
elapsed_time = (end_time - start_time).total_seconds()

File: snafu/log_generator_wrapper/trigger_log_generator.py [hunk 2]
```diff
@@ -0,0 +1,274 @@
...
+        message_count = self._run_log_test()
+        end_time = time.time()
+        elaspsed_time = end_time - start_time
```
rsevilla87, line 228:
```suggestion
        elapsed_time = end_time - start_time
```

dry923: fixed

snafu/log_generator_wrapper/trigger_log_generator.py
dry923: @rsevilla87 fair point. It functionally works out the same though in terms of the end result. I'll play around with it after the demo today since I don't want to break it prior :-D

rsevilla87:
@dry923 I wonder if we should execute all pod synchronization from here rather than benchmark-operator.

dry923:
> @dry923 I wonder if we should execute all pod synchronization from here rather than benchmark-operator. Looking at other more complex cases like uperf I think we should standardize the place where we perform this kind of operations.

@rsevilla87 updated the operator PR to include managing the syncing of the pods (see: https://github.com/dry923/ripsaw/blob/da1b75beafd66f0b9265bdaca97118923fc31c8c/roles/log_generator/templates/log_generator.yml#L75). Do you think its worth leaving the code in here as well in case it is used outside the operator?

dry923:
> > @dry923 I wonder if we should execute all pod synchronization from here rather than benchmark-operator. Looking at other more complex cases like uperf I think we should standardize the place where we perform this kind of operations.
> 
> @rsevilla87 updated the operator PR to include managing the syncing of the pods (see: https://github.com/dry923/ripsaw/blob/da1b75beafd66f0b9265bdaca97118923fc31c8c/roles/log_generator/templates/log_generator.yml#L75). Do you think its worth leaving the code in here as well in case it is used outside the operator?

@rsevilla87 removed code from the wrapper

jtaleric (approved):
LGTM!

rsevilla87 (approved):
LGTM!

PR #255, opened 2021-02-24
Set the desired upgrade version dynamically

chaitanyaenr:
This commit enables to get the desired version dynamically during
the upgrade run and compares it with the initial version before
the cluster upgrade has been triggered. This way, we don't need to
pass the version option.


snafu/upgrade_openshift_wrapper/trigger_upgrade.py
```diff
@@ -25,7 +25,6 @@ class Trigger_upgrade():
     def __init__(self, args):
         self.uuid = args.uuid
         self.user = args.user
-        self.version = args.version
```
rsevilla87, line 28: I think we still need this variable,  `self.version` is referenced in the line 92, to trigger an upgrade to a specific version

rsevilla87 (approved):
LGTM!

jtaleric (approved):
LGTM!

PR #256, opened 2021-03-09
Fixing Scale Kubernetes Client Configuration

whitleykeith:
This is due to https://github.com/kubernetes-client/python/issues/1284 which is present in the Python Kubernetes Client with version v12 and higher. Looks like the latest version of the openshift client now pulls in a v12 K8s client which is causing this problem. 

This comment https://github.com/kubernetes-client/python/issues/1284#issuecomment-750737547 recommends this fix. I also added some exception handling so that the process doesn't exit 0 when it actually failed


chaitanyaenr (approved):
LGTM

PR #258, opened 2021-03-16
Add delay before checking for desired upgrade version

### LINKED ISSUE #257: [Upgrade workload] Desired version takes a couple of seconds to switch
The desired version field of the clusterversion takes 2-3 seconds to change after an upgrade has been triggered. This is leading to issues with capturing the current/desired versions, introducing a delay/sleep 10 before grabbing the desired version for comparison should fix it.

chaitanyaenr:
This is needed as it takes a couple of seconds for the desired version
in the clusterversion spec to get updated after an upgrade has been
triggered. This commit also enables upgrading with force and even when
there are warnings around image verification.

Fixes #257

### DIFF
--- a/snafu/upgrade_openshift_wrapper/trigger_upgrade.py
+++ b/snafu/upgrade_openshift_wrapper/trigger_upgrade.py
@@ -84,7 +84,7 @@ class Trigger_upgrade():
         # Upgrade to the latest build available in the channel if set
         elif self.latest:
             cmd = (
-                "oc adm upgrade --to-latest=true")
+                "oc adm upgrade --to-latest=true --allow-upgrade-with-warnings=true --force=true")
         elif self.version:
             cmd = (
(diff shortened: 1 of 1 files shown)


chaitanyaenr:
@jtaleric We had to use --allow-upgrade-with-warnings flag since cvo runs the upgrades only when there are no warnings and looks like the nightly images are not verified thus generating a warning which is blocking the upgrades. Using this flag can potentially override the on-going upgrade, maybe we should add in a check to track that and only upgrade when there isn't one in progress, will track the issue and create a different PR to cover that case.

jtaleric (commented):
Is it possible to remove the sleep and just check for the version?

chaitanyaenr:
@jtaleric yes, updated the PR to check for the version. PTAL.

jtaleric (approved):
LGTM

PR #267, opened 2021-04-29
Add DNS performance workload wrapper

chaitanyaenr:
### Description
This commit adds a wrapper to run dns-perf which sends queries to
the specified DNS server given a profile with the records in order
to test the DNS performance. It parses the results and generated a
 json payload which can be indexed into Elasticsearch to help with
 the analysis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ environments, develop in SNAFU and then write benchmark-operator benchmark to in
 | Log Generator                  | Log throughput to backend      | Working            |
 | Image Pull                     | Time to copy from a container image from a repo | Working |
 | sysbench                       | CPU,Memory,Mutex,Threads,Fileio | Working |
+| DNS-Perf                       | DNS Performance                | Working            |
 
 ## What backend storage do we support?
 
--- a/snafu/dns_perf_wrapper/Dockerfile
+++ b/snafu/dns_perf_wrapper/Dockerfile
@@ -1,0 +1,11 @@
+FROM registry.centos.org/centos:8
+
+COPY snafu/image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
+RUN dnf install -y epel-release && dnf install -y --nodocs redis python3 python3-pip openssl-devel ck-devel yum-plugin-copr && \
+    dnf copr enable @dnsoarc/dnsperf -y && \
+    dnf install dnsperf -y && \
+    dnf clean all && rm -rf /var/cache/yum
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz | tar xz -C /usr/bin/ oc && ln -s /usr/bin/python3 /usr/bin/python
+RUN mkdir -p /opt/snafu/
+COPY . /opt/snafu/
+RUN pip3 install -e /opt/snafu/
--- a/snafu/dns_perf_wrapper/Readme.md
+++ b/snafu/dns_perf_wrapper/Readme.md
@@ -1,0 +1,88 @@
+## DNS Performance Workload Wrapper
+Wrapper around [dnsperf](https://github.com/DNS-OARC/dnsperf) to measure throughput ( requests per second ) and latency metrics for Domain Name Service ( DNS ). The collected metrics are scraped to generate a json which can be indexed into Elasticsearch datastore.
+
+
+### Dependencies
+This wrapper requires a couple of dependencies as well as enabling copr repo to download the latest dnsperf package. For more details [see[(https://github.com/DNS-OARC/dnsperf/blob/master/README.md). It's recommended to build and run the [containerized version](Dockerfile) to avoid managing the installation on the host. Here are the dependencies which need to be installed when running the standalone version:
+- python3
+- python3-pip
+- openssl-devel
+- epel-release
+- ck-devel
+- yum-plugin-copr
+- dnsperf ( Run $ dnf copr enable @dnsoarc/dnsperf and $ dnf install dnsperf to enable the repo to install it )
+
+
+### Run
+```
+usage: run_snafu [-h] [-v] -t TOOL [--run-id [RUN_ID]] -u UUID --server-address SERVER_ADDRESS --queries-per-second QUERIES_PER_SECOND --run-time RUN_TIME --data-file DATA_FILE [--clients CLIENTS]
+
+DNS perf workload Wrapper script
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -v, --verbose         enables verbose wrapper debugging info (default: 20)
+  -t TOOL, --tool TOOL  Provide tool name (default: None)
+  --run-id [RUN_ID]     Run ID to unify benchmark results in ES (default: NA)
+
+DNS Perf:
+  -u UUID, --uuid UUID  Provide the uuid (default: None)
+  --server-address SERVER_ADDRESS
+                        DNS server address to send the requests (default: None)
+  --queries-per-second QUERIES_PER_SECOND
+                        Number of queries per second (default: None)
+  --run-time RUN_TIME   run for at most this many seconds (default: None)
+  --data-file DATA_FILE
+                        File path with the records to send as queries to the server (default: None)
+  --clients CLIENTS     The number of clients to act as (default: 1)
+```
+
+Note that file location passed through --data-file variable is where the records are defined. For example:
+```
+www.google.com A
+www.gmail.com A
+chat.google.com A
+mail.google.com A
+drive.google.com A
+docs.google.com A
+meet.google.com A
+sites.google.com A
+plus.google.com A
+```
+
+### Metrics
+The wrapper runs dnsperf and generates a json document with the results as well as the input configuration/parameters. Benchmark wrapper also wraps around metadata to be indexed into Elasticsearch as explained [here](https://github.com/cloud-bulldozer/benchmark-wrapper#how-do-i-post-results-to-elasticsearch-from-my-wrapper). Here is a sample results document:
+
+```
+{
+    "_index": "snafu-dns_perf-results",
+    "_op_type": "create",
+    "_source": {
+        "uuid": "0f038d28-ef3a-42c0-aea3-435d19686ab1",
+        "cluster_name": "mycluster",
+        "clients": 1,
+        "timestamp": "2021-05-21T16:45:15",
+        "elapsed_time": 6.012354850769043,
+        "latency_stddev": 0.003623,
+        "avg_latency": 0.010767,
+        "latency_min": 0.006275,
+        "latency_max": 0.017417,
+        "QPS": 19.999643,
+        "runtime": 6.000107,
+        "avg_request_packet_size": 32.0,
+        "avg_response_packet_size": 60.0,
+        "queries_lost": 0,
+        "queries_lost_percentage": 0.0,
+        "queries_completed": 120,
+        "queries_completed_percentage": 100.0,
+        "queries_sent": 120,
+        "stop_time_duration": 6.0,
+        "dns_address": "8.8.8.8:53",
+        "dnsperf_version": "2.5.2",
+        "run_id": "NA"
+    },
+    "_id": "18cdda19f6d15c6dab17c6da7fa1422f007057a193b25d298439225c0325ce8f",
+    "run_id": "NA"
+}
+
+```
--- a/snafu/dns_perf_wrapper/trigger_dns_perf.py
+++ b/snafu/dns_perf_wrapper/trigger_dns_perf.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+import time
+import logging
+import subprocess
+from ttp import ttp
+
+logger = logging.getLogger("snafu")
+
+
+class Trigger_dns_perf:
+    def __init__(self, args):
+        """ Initialize the arguments defined """
+
+        self.uuid = args.uuid
+        self.cluster_name = args.cluster_name
+        self.server_address = args.server_address
+        self.queries_per_second = args.queries_per_second
+        self.run_time = args.run_time
+        self.data_file = args.data_file
+        self.clients = args.clients
+
+    def _json_payload(self, data, timestamp, elapsed_time):
+        """ Generates a json payload with the metrics of interest.
+        This is the results document, it gets updated with metrics from the dns-perf run.
+        """
+
+        payload = {
+            "uuid": self.uuid,
+            "cluster_name": self.cluster_name,
+            "clients": self.clients,
+            "timestamp": timestamp,
+            "elapsed_time": elapsed_time,
+        }
+        try:
+            payload.update(data)
+        except Exception as e:
+            logging.error("Failed to update the result document, error: %s" % (e))
+        return payload
+
+    def run(self):
+        """ Runs dns-perf workload and returns the stdout """
+
+        command = "dnsperf -l {} -s {} -Q {} -d {} -c {}".format(
+            self.run_time, self.server_address, self.queries_per_second, self.data_file, self.clients
+        )
+        logger.info(command)
+        out = subprocess.Popen(
+            command, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+        )
+        (output, err) = out.communicate()
+        if out.returncode != 0:
+            logger.error("Failed to run %s, error: %s" % (command, err))
+            exit(1)
+        logger.info("Raw result: \n{}".format(output))
+        return output
+
+    def _parse_stdout(self, stdout):
+        """ Parses the dns-perf run stdout.
+            ttp template to parse the dns-perf stdout.
+        """
+
+        template = """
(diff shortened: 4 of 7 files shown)


README.md
```diff
@@ -24,7 +24,8 @@ environments, develop in SNAFU and then write benchmark-operator benchmark to in
 | OpenShift Upgrade              | Time to upgrade                | Working            |
 | OpenShift Scaling              | Time to scale                  | Working            |
 | Log Generator                  | Log throughput to backend      | Working            |
-
+| DNS-Perf                       | DNS Performance                | Working            |
```
chaitanyaenr, line 27: Add documentation around the wrapper - what the workload is and the metrics captured.

snafu/dns_perf_wrapper/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.centos.org/centos:8
+
+COPY snafu/image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
+RUN dnf install -y --nodocs python3 python3-pip epel-release && \
```
jtaleric, line 4: Can we just switch to a single `dnf install`

snafu/dns_perf_wrapper/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+    dnf install -y openssl-devel ck-devel yum-plugin-copr && \
+    dnf copr enable @dnsoarc/dnsperf -y && \
+    dnf install dnsperf -y && \
+    dnf clean all && rm -rf /var/cache/yum
+RUN dnf install -y --nodocs redis --enablerepo=centos8-appstream && dnf clean all
```
jtaleric, line 9: This should be moved above the `install / clean ` to reduce redundancy.

snafu/dns_perf_wrapper/Readme.md
```diff
@@ -0,0 +1,39 @@
+## DNS Performance Workload Wrapper
+Wrapperf around [dnsperf](https://github.com/DNS-OARC/dnsperf) to measure throughput ( requests per second ) and latency metrics for Domain Name Service ( DNS ). The collected metrics are scraped to generate a json which can be indexed into Elasticsearch datastore.
```
jtaleric, line 2: `Wrapperf`. what is latency measured in?

snafu/dns_perf_wrapper/Readme.md
```diff
@@ -0,0 +1,39 @@
...
+        "latency_stddev": "0.003210",
+        "avg_latency": "0.012043",
+        "QPS": "19.999700",
+        "runtime": "6.000090",
+        "avg_request_packet_size": "32",
+        "avg_response_packet_size": "60",
+        "queries_lost": "0",
+        "queries_lost_percentage": "0.00%",
+        "queries_completed": "120",
+        "queries_completed_percentage": "100.00%",
```
jtaleric, line 29: Removing the `%`, the key states it is percentage.

snafu/dns_perf_wrapper/Readme.md
```diff
@@ -0,0 +1,39 @@
...
+        "timestamp": "2021-04-29T19:17:47",
+        "elapsed_time": 6.0132739543914795,
+        "latency_stddev": "0.003210",
+        "avg_latency": "0.012043",
+        "QPS": "19.999700",
+        "runtime": "6.000090",
+        "avg_request_packet_size": "32",
+        "avg_response_packet_size": "60",
+        "queries_lost": "0",
+        "queries_lost_percentage": "0.00%",
```
jtaleric, line 27: Removing the `%`, the key states it is percentage.

snafu/dns_perf_wrapper/Readme.md
```diff
@@ -0,0 +1,39 @@
...
+This wrapper requires a couple of dependencies as well as enabling copr repo to download the latest dnsperf package. For more details [see[(https://github.com/DNS-OARC/dnsperf/blob/master/README.md). It's recommended to build and run the [containerized version](Dockerfile) to avoid managing the installation on the host.
+
+### Metrics
+The wrapper runs dnsperf and generates a json document with the results as well as the input configuration/parameters. Benchmark wrapper also wraps around metadata to be indexed into Elasticsearch as explained [here](https://github.com/cloud-bulldozer/benchmark-wrapper#how-do-i-post-results-to-elasticsearch-from-my-wrapper). Here is a sample results document:
+
+```
+{
+    "_index": "snafu-dns_perf-results",
+    "_op_type": "create",
+    "_source": {
```
jtaleric, line 14: Should some of these values be `ints`? Not all strings...

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+        self.server_address = args.server-address
+        self.queries_per_second = args.queries-per-second
+        self.run_time = args.run-time
+        self.data_file = args.data-file
+        self.clients = args.clients
+
+    # Generate a json payload with the metrics of interest
+    # This is the results document, it gets updated with metrics from the dns-perf run
+    def _json_payload(self, data, timestamp, elapsed_time):
+        #
```
jtaleric, line 39: Random comment?

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+from ttp import ttp
+
+logger = logging.getLogger("snafu")
+
+class Trigger_dns_perf():
+    # Initialize the arguments defined
+    def __init__(self, args):
+        self.uuid = args.uuid
+        self.cluster_name = args.cluster_name
+        self.server_address = args.server-address
```
jtaleric, line 30: This works? Example of my concern 
```
>>> test-var=10
  File "<stdin>", line 1
    test-var=10
    ^
SyntaxError: cannot assign to operator
>>> test_var=10
>>> 

```

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+        return payload
+
+    # Run the dns-perf workload
+    def run_dns_perf_test(self):
+        command = "dnsperf -l {} -s {} -Q {} -d {} -c {}".format(self.run_time, self.server_address, self.queries_per_second, self.data_file, self.clients)
+        logger.info(command)
+        try:
+            out = subprocess.Popen(command, shell=True,
+                                   universal_newlines=True, stdout=subprocess.PIPE)
+            (output, err) = out.communicate()
```
jtaleric, line 60: Set `err` but never do anything with it?

jtaleric (changes requested):
some nits and questions

snafu/dns_perf_wrapper/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/latest/openshift-client-linux.tar.gz | tar xz -C /usr/bin/ oc && ln -s /usr/bin/python3 /usr/bin/python
```
kurtismullins, line 10: Why is python3 symlinked to python here?

README.md
chaitanyaenr: Done, added documentation around the wrapper.

snafu/dns_perf_wrapper/Readme.md
```diff
@@ -0,0 +1,39 @@
+## DNS Performance Workload Wrapper
+Wrapperf around [dnsperf](https://github.com/DNS-OARC/dnsperf) to measure throughput ( requests per second ) and latency metrics for Domain Name Service ( DNS ). The collected metrics are scraped to generate a json which can be indexed into Elasticsearch datastore.
+
+### Dependencies
+This wrapper requires a couple of dependencies as well as enabling copr repo to download the latest dnsperf package. For more details [see[(https://github.com/DNS-OARC/dnsperf/blob/master/README.md). It's recommended to build and run the [containerized version](Dockerfile) to avoid managing the installation on the host.
```
kurtismullins, line 5: The dependencies section is confusing me a bit. It might be useful to exclude mention of the copr repo as that's not the dependency, but where the dependency is currently being retrieved from (if I am following correctly)

I'm not sure if this section is noting the the dependencies of the constructed Dockerfile, or the wrapper. If it's the latter case, something like this might be a little easier to follow (and maintain):

- dnsperf
- python3 (may want to specify a specific version of Python, e.g. 3.8)
- etc...

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+            "timestamp": timestamp,
+            "elapsed_time": elapsed_time
+        }
+        try:
+            payload.update(data)
+        except Exception as e:
+            logging.error("Failed to update the result document, error: %s" %(e))
+        return payload
+
+    # Run the dns-perf workload
```
kurtismullins, line 53: (nitpick) I would move this to a docstring within the function, itself, and explain the purpose/usage of it. Example:

```
run_dns_perf_test(self):
    """
    Execute a dnsperf process until it finishes and return the stdout.
    """
    ...
```

Related -- this is using a lot of `self` referenced variables. It might make more sense to simply call this `run()` or `execute_process()` since it would be invoked along the lines of:

```
self.run()
# or 
dns_perf_test.run()
```

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+            out = subprocess.Popen(command, shell=True,
+                                   universal_newlines=True, stdout=subprocess.PIPE)
+            (output, err) = out.communicate()
+            logger.info("Raw result: \n{}".format(output))
+        except Exception as e:
+            logger.error("Failed to run %s, error: %s" % (command, e))
+            exit(1)
+        return output
+
+    # Parse the output generated by dns-perf workload
```
kurtismullins, line 67: (nitpick) docstring within the function instead of a comment above the function

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+  Average packet size:  request {{ avg_request_packet_size }}, response {{ avg_response_packet_size }}
+  Run time (s):         {{ runtime }}
+  Queries per second:   {{ QPS }}
+
+  Average Latency (s):  {{ avg_latency }} (min 0.008983, max 0.017859)
+  Latency StdDev (s):   {{ latency_stddev }}
+        """
+       parser = ttp(stdout, template)
+       parser.parse()
+       parsed_results = parser.result()[0][0]
```
kurtismullins, line 96: I am unfamiliar with `ttp`. Is there a reason you are using it to generate a single string? You could probably get the same results with a dictionary and using `.format()` on your template.

https://docs.python.org/3/library/stdtypes.html#str.format

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,117 @@
...
+       logger.info("Parsed results:")
+       logger.info(parsed_results)
```
kurtismullins, line 98: This is something I am struggling with in a different test. On one-hand, it's useful to dump the data out where it will be read. On the other hand, most standard logs are single-line statements unless an error occurred and you're logging a stack trace. No change needed here -- just something to ponder.

kurtismullins (changes requested):
I believe it's a great start. I have left some comments that may be more in the style/nitpick spectrum of priority :)

Waiting to see some important items mentioned by @jtaleric addressed before signing off.

snafu/dns_perf_wrapper/Dockerfile
chaitanyaenr: we need epel-release to install a couple of other dependencies which is why we had to go with multiple dnf commands.

snafu/dns_perf_wrapper/Dockerfile
chaitanyaenr: Fixed it, thanks.

snafu/dns_perf_wrapper/Readme.md
chaitanyaenr: Fixed it, thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: Fixed it, thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: Right, modified to avoid python interpreting hyphens as subtraction operator.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: Fixed it, thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: Nice suggestion, fixed it. Thanks.

snafu/dns_perf_wrapper/Readme.md
chaitanyaenr: You are right @kurtismullins, copr repo is where the dependency is being retrieved from but it's a custom repo needed to download dns-perf, so it kind of acts as a dependency :-)

Will mention the dependencies which are mentioned in the dockerfile in the readme to make it clear like you suggested. Thanks.

jtaleric:
@chaitanyaenr checking in on this?

snafu/dns_perf_wrapper/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.centos.org/centos:8
+
+COPY snafu/image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
+RUN dnf install -y epel-release && dnf install -y --nodocs redis python3 python3-pip openssl-devel ck-devel yum-plugin-copr && \
```
jtaleric, line 4: Why do we have 3 dnf commands? Can you reorder and put `dnf copr enable` as the top most, and intall all the necessary packages in a single command?

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,125 @@
...
+            self.run_time, self.server_address, self.queries_per_second, self.data_file, self.clients
+        )
+        logger.info(command)
+        try:
+            out = subprocess.Popen(command, shell=True, universal_newlines=True, stdout=subprocess.PIPE)
+            (output, err) = out.communicate()
+            logger.info("Raw result: \n{}".format(output))
+        except Exception as e:
+            logger.error("Failed to run %s, error: %s" % (command, e))
+            logger.error(err)
```
jtaleric, line 65: Not sure if err will be in scope here? @kurtismullins ?

I think output == `stdout` and err == `stderr` - https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate 

Which shouldn't fire up an exception... Just if the command errors out/returns non-zero

snafu/dns_perf_wrapper/Dockerfile
chaitanyaenr: We symlink it to make sure python3 is used as the interpreter.

snafu/dns_perf_wrapper/Dockerfile
chaitanyaenr: Discussed about this in slack. Thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: You are right, updated it act on returncode. Thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
chaitanyaenr: Fixed it, thanks.

chaitanyaenr:
@kurtismullins @dry923 made the changes requested, PTAL. Thanks.

snafu/dns_perf_wrapper/Readme.md
chaitanyaenr: Right, converted the ttp parsed data to int's, float's and strings respectively. Thanks.

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,125 @@
...
+
+        payload = {
+            "uuid": self.uuid,
+            "cluster_name": self.cluster_name,
+            "clients": self.clients,
+            "timestamp": timestamp,
+            "elapsed_time": elapsed_time,
+        }
+        try:
+            payload.update(data)
```
kurtismullins, line 47: This is a clean approach!

snafu/dns_perf_wrapper/trigger_dns_perf.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def run(self):
+        """ Runs dns-perf workload and returns the stdout """
+
+        command = "dnsperf -l {} -s {} -Q {} -d {} -c {}".format(
+            self.run_time, self.server_address, self.queries_per_second, self.data_file, self.clients
+        )
+        logger.info(command)
+        out = subprocess.Popen(
+            command, shell=True, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE
```
kurtismullins, line 60: Just curious, is there a reason you're using shell here? It has security implications.

snafu/utils/wrapper_factory.py
```diff
@@ -16,7 +16,11 @@
 from snafu.flent_wrapper.flent_wrapper import flent_wrapper
 from snafu.log_generator_wrapper.log_generator_wrapper import log_generator_wrapper
 from snafu.image_pull_wrapper.image_pull_wrapper import image_pull_wrapper
+<<<<<<< HEAD
```
jtaleric, line 19: yikes

jtaleric (approved):
LGTM @kurtismullins ?

kurtismullins (approved):
LGTM

PR #269, opened 2021-05-07
fix ceph cachedrop URL and env. var. name

bengland2:
### Description

correct the URL used to invoke the cache dropper
use env. var. for cache drop pod that matches benchmark-operator
depends on recently merged [bohica PR 19](https://github.com/cloud-bulldozer/bohica/pull/19)

### Fixes

need to specify Ceph cache dropper pod IP in workload CR.  Now benchmark-operator will get it and pass it to wrapper.


jtaleric (approved):
LGTM

PR #272, opened 2021-05-18
Use different group_ids per log generator pod

smalleni:


dry923 (approved):
LGTM!

PR #273, opened 2021-05-19
Disable SNAFU logging

smalleni:
This is a corner case needed during our log testing.
To accurately count the number of synthetic log messages
generated by the log generator pod, we want to disable the regular SNAFU
logging. This will make it harder to debug, but this is only something
a user optionally sets in the benchmark-operator CR. The reason we want
to do this is, we plan to deploy only the log-generator pods to a separate
namespace and forward those logs to a separate kafka topic and then we can
reliably read off the offset in the topic without reading each message to
verify that all log messages generated by the log generator pod made it to kafka.

### DIFF
--- a/snafu/log_generator_wrapper/trigger_log_generator.py
+++ b/snafu/log_generator_wrapper/trigger_log_generator.py
@@ -22,9 +22,12 @@ import boto3
 import json
 import subprocess
 import uuid
+import os
 from kafka import KafkaConsumer, TopicPartition
 
 logger = logging.getLogger("snafu")
+if (os.getenv("snafu_disable_logs", "False")) == "True":
+    logger.disabled = True
 
 
 class Trigger_log_generator:


snafu/log_generator_wrapper/trigger_log_generator.py
```diff
@@ -22,9 +22,12 @@
 import json
 import subprocess
 import uuid
+import os
 from kafka import KafkaConsumer, TopicPartition
 
 logger = logging.getLogger("snafu")
+if (os.getenv("snafu_disable_logs", "False")) == "True":
```
dry923, line 29: Can we make this a parameter to the wrapper instead of a Env var?

smalleni: It shouldn't matter which way it is consumed right? Is there a reason for this preference? I would much rather have it as an env var because I can use it within the pod's command/args to quiesce any bash echos as well.

dry923: Predominately it is to keep consistent with the other options that are in use in the wrapper but if it is going to hinder your workflow its fine.

smalleni: > Predominately it is to keep consistent with the other options that are in use in the wrapper but if it is going to hinder your workflow its fine.

Great, thanks!

smalleni: I don't see many users doing this at all and we also have a sane default. So perhaps not adding to the args at all, was my take. Also, to be able to use it inside the bash script in the pod.

dry923 (approved):
LGTM!

PR #276, opened 2021-05-24
Improvements to kafka checks

smalleni:
- Update backend details to the doc sent to ES
- Close kafka consumer
- Increase the `end_time` used to get the kafka offset
  This is because, unlike ES, the timestamp is the one created
  by Kafka when the log message is recorded in Kafka. For example,
  in the case of ES, the timestamp added by FluenD becomes the timestamp
  field in ES, so we can query accunrately for log messages between the
  `start_time` and `end_time` of the log generation test. However, in the case
  of Kafka, we cannot accurately check for the log messages generated during
  the test because the records in Kafka and their corresponding timstamps for
  the offsets could significantly after the `end_time` of the test as the messages
  slowly show up in the kafka cluster and there is no way to use the timestamp from
  FluentD as the timstamp for the kafka offset and the FluentD timestamp is actually
  in the content of the Kafka record. We currently check for all messages from the
  `start_time` and before the `timeout` configured in the CR.

### DIFF
--- a/snafu/log_generator_wrapper/trigger_log_generator.py
+++ b/snafu/log_generator_wrapper/trigger_log_generator.py
@@ -92,6 +92,7 @@ class Trigger_log_generator:
                 "kafka_topic": self.kafka_topic,
                 "backend": "kafka",
             }
+            payload.update(backend)
         payload.update(data)
         return payload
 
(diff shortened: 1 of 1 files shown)


dry923 (approved):
LGTM!

PR #277, opened 2021-05-25
Minor refactoring

smalleni:

### DIFF
--- a/snafu/log_generator_wrapper/log_generator_wrapper.py
+++ b/snafu/log_generator_wrapper/log_generator_wrapper.py
@@ -59,6 +59,7 @@ class log_generator_wrapper:
         )
         parser.add_argument("--kafka-bootstrap-server", help="The Kafka Service to connect to")
         parser.add_argument("--kafka-topic", help="The Kafka topic to verify logs")
+        parser.add_argument("--kafka-check", help="Verify if logs made it to kafka sink", action="store_true")
         self.args = parser_object.parse_args()
 
         self.args.cluster_name = os.getenv("clustername", "mycluster")
--- a/snafu/log_generator_wrapper/trigger_log_generator.py
+++ b/snafu/log_generator_wrapper/trigger_log_generator.py
@@ -51,13 +51,13 @@ class Trigger_log_generator:
         self.es_index = args.es_index
         self.kafka_bootstrap_server = args.kafka_bootstrap_server
         self.kafka_topic = args.kafka_topic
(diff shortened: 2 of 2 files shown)


smalleni:
@dry923 this is tested locally and ready for review. Improves some logging, moves around some code to make it cleaner.

snafu/log_generator_wrapper/trigger_log_generator.py
```diff
@@ -276,15 +282,17 @@ def emit_actions(self):
...
-                    logger.info(
-                        "Current messages received is {}, waiting more time for kafka".format(
-                            messages_received
-                        )
-                    )
                 if messages_received == message_count:
                     received_all_messages = True
                 else:
                     logger.info("Message check failed. Retrying until timeout")
+                    if self.kafka_bootstrap_server:
```
dry923, line 289: This if is redundant with the if at line 274. It wouldn't fall into here if the bootstrap server was undefined

snafu/log_generator_wrapper/log_generator_wrapper.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, parent_parser):
         )
         parser.add_argument("--kafka-bootstrap-server", help="The Kafka Service to connect to")
         parser.add_argument("--kafka-topic", help="The Kafka topic to verify logs")
+        parser.add_argument("--kafka-check", help="Verify if logs made it to kafka sink", action="store_true")
```
dry923, line 62: @smalleni Curious why you are adding a verification true|false. If you are providing the topic/bootstrap then the verification would be implied. Or is there another time/reason you would pass that info but not want to verify?

snafu/log_generator_wrapper/trigger_log_generator.py
smalleni: It is not, I only want to execute L290-L295 for kafka. Line 274 is an OR condition, so we enter into it for any condition that meets it (ES, cloudwatch or kafka), however I now only want to check for kafka.

dry923: Ahhhhhhhhh ok. nevermind.

snafu/log_generator_wrapper/log_generator_wrapper.py
dry923: That's fair.  I just wanted to confirm the intent really.

jtaleric (approved):
lgtm

PR #279, opened 2021-05-26
Added functionality for creating and processing archive file

acalhounRH:
### Description
added functionality to create and re-index a archive file for benchmark-wrapper
when creating an archive we would just include --create-archive on any run and it will drop a `user_clustername_uuid.archive` file in the cwd

for example `python3.7 ~/git/benchmark-wrapper/snafu/run_snafu.py --tool sysbench -f /home/acalhoun/sysbench_results/example_cpu_test.conf --create-archive`

when indexing from an archive we would do, `run_snafu --tool archive --archive-file <file>`. run_snafu will go straight to processing and indexing the listed ES friendly documents in the archive.

For example `python3.7 ~/git/benchmark-wrapper/snafu/run_snafu.py --tool archive --archive-file myuser_mycluster_1b88311d-1230-409a-9ffa-94e6f9aa4754.archive -v`

@jtaleric @dry923

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,74 @@
 # Benchmark-Wrapper aka SNAFU - Situation Normal: All F'ed Up
 
-Most Performance workload tools were written to tell you the performance at a given time under given circumstances.
+Benchmark-wrapper provides a convenient mechanism for launching, processing, and storing data produced by a suite of performance benchmarks. Users can run Benchmark-wrapper in a traditional bare-metal environment or with the use of [benchmark-operator](https://github.com/cloud-bulldozer/benchmark-operator) ran in a containerized environment such as Kubernetes.
 
-These scripts are to help enable these legacy tools store their data for long term investigations.
+Note: If you need your benchmark to collect data for both Kubernetes and non-Kubernetes
+environments, incorporate your benchmark into benchmark-wrapper and then write a benchmark-operator benchmark to integrate with Kubernetes.
 
-Note: SNAFU does not depend upon Kubernetes, so you can use run_snafu.py on a bare-metal or VM cluster without relying
-on Kubernetes to start/stop pods.  So if you need your benchmark to collect data for both Kubernetes and non-Kubernetes
-environments, develop in SNAFU and then write benchmark-operator benchmark to integrate with Kubernetes.
+**Why Should I use Benchmark-wrapper?**
+
+Traditionally benchmark tools have presented users with an adhoc raw standard output, limiting ability to perform detailed statistical analysis, no way to preserve results for long term archive, and difficulty at being platform agnostic. Benchmark-wrapper aims to solve all of these issues and provide users with a integrated streamlined interface.
+
+# How to Run
+
+
+It is suggested to use a virtual environment to install and run snafu.
+
+```
+python3 -m venv /path/to/new/virtual/environment
+source /path/to/new/virtual/environment/bin/activate
+git clone https://github.com/cloud-bulldozer/snafu
+python setup.py develop
+run_snafu --tool Your_Benchmark ...
+```
+## - Install ##
+
+```
+git clone https://github.com/cloud-bulldozer/benchmark-wrapper.git
+sudo pip3 install /path-to-benchmark-wrapper/benchmark-wrapper
+```
+
+## - Configure ##
+Benchmark-wrapper uses several environment variable to provide user context and interfaces to ES.
+
+```
+export uuid=<RFC4122 Version 4 uuid>
+export test_user=<test user name>
+export clustername=<platform or cluster descriptive name>
+export es=<http://es_address:es_port>
+```
+
+## - Run ##
+
+```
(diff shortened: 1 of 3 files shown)


dry923:
@acalhounRH can you make it so the archive file location and possibly name can be set?

acalhounRH:
Sure thing I used the --archive-file=<path to file/filename> and the user can provide a full path to file, it will default to cwd and user_cluster_uuid.archive.

snafu/run_snafu.py
```diff
@@ -99,9 +108,23 @@ def main():
     # call py es bulk using a process generator to feed it ES documents
     if index_args.index_results:
         parallel_setting = strtobool(os.environ.get("parallel", "false"))
-        res_beg, res_end, res_suc, res_dup, res_fail, res_retry = streaming_bulk(
-            es, process_generator(index_args, parser), parallel_setting
-        )
+
+        if "archive" in index_args.tool:
+            if index_args.archive_file:
```
dry923, line 113: @acalhounRH doesn't there need to be an if !create-archive as well with this? Otherwise if you are trying to create one w/ a specific name/dest it will try to read it instead. Or am I just horribly misreading the code :-D

acalhounRH: Using your own word, just horribly misreading the code :-D

If --create-archive
    create the archive file, if user provided file use that if not then use the default scheme
else: # user did not provide --create-archive 
   do your normal thing

jtaleric (commented):
Minor nit, but can we add a test to archive then check if the archive is present?

acalhounRH:
> Minor nit, but can we add a test to archive then check if the archive is present?

No testing within benchmark-wrapper, would this request be a follow on activity for benchmark-operator?

snafu/run_snafu.py
dry923: ack what i wasn't grasping was that you actually set a tool called archive for this. Makes sense now.

dry923:
@acalhounRH can you update the Readme for this as well?

jtaleric:
> > Minor nit, but can we add a test to archive then check if the archive is present?
> 
> No testing within benchmark-wrapper, would this request be a follow on activity for benchmark-operator?

bah-humbug...

README.md
```diff
@@ -1,12 +1,76 @@
...
+
+for example: 
+
+```
+python3.7 ./snafu/run_snafu.py --tool sysbench -f example__cpu_test.conf
+```
+
+## Archiving data
+
+Benchmark-wrapper has two forms of capturing data. The first and preferred method is directly writing data to Elasticsearch, users will need to set the **es** environment variable in order to enable this. The second method
```
dry923, line 56: You may want to point out that these are not mutually exclusive. i.e. you can index to ES and write to the archive on the same run.

acalhounRH: added

README.md
```diff
@@ -1,12 +1,74 @@
...
+
+for example:
+
+```
+python3.7 ./snafu/run_snafu.py --tool sysbench -f example__cpu_test.conf
+```
+
+## Archiving data
+
+Benchmark-wrapper has two forms of capturing data. The first and preferred method is directly writing data to Elasticsearch, users will need to set the **es** environment variable in order to enable this. The second method used for capturing data is writing to a local archive file, this is intended to be enabled when elasticsearch is not available for direct indexing or for a backup of indexed results. both methods can be enabled at the same time, and are independent of each other.
```
dry923, line 55: Capital B in both for the last sentence.

acalhounRH: done

dry923 (approved):
LGTM pending the last CI passing.

README.md
```diff
@@ -1,12 +1,74 @@
...
+Benchmark-wrapper provides a convenient mechanism for launching, processing, and storing data produced by a suite of performance benchmarks. Users can run Benchmark-wrapper in a traditional bare-metal environment or with the use of [benchmark-operator](https://github.com/cloud-bulldozer/benchmark-operator) ran in a containerized environment such as Kubernetes.
 
-These scripts are to help enable these legacy tools store their data for long term investigations.
+Note: If you need your benchmark to collect data for both Kubernetes and non-Kubernetes
+environments, incorporate your benchmark into benchmark-wrapper and then write a benchmark-operator benchmark to integrate with Kubernetes.
 
-Note: SNAFU does not depend upon Kubernetes, so you can use run_snafu.py on a bare-metal or VM cluster without relying
-on Kubernetes to start/stop pods.  So if you need your benchmark to collect data for both Kubernetes and non-Kubernetes
-environments, develop in SNAFU and then write benchmark-operator benchmark to integrate with Kubernetes.
+** Why Should I use Benchmark-wrapper? **
```
jtaleric, line 8: Did you expect some formatting like h2 with this?

acalhounRH: Was expecting Bold

jtaleric: ah, i don't think it worked.

jtaleric (approved):
LGTM

PR #281, opened 2021-06-07
Correct license in setup.cfg (mit -> apache2)

learnitall:
### Description

Correct the license declared in setup.cfg. Was set to MIT when it should be set to Apache 2.0

For reference, set the license string by example from https://github.com/apache/airflow/blob/main/setup.cfg, which also uses Apache 2.0


jtaleric (approved):
LGTM!

PR #282, opened 2021-06-10
Update gitignore with more Python and IDE files.

learnitall:
### Description

Adds more Python files and IDE settings files to the gitignore. Based on the following:

* https://djangowaves.com/tips-tricks/gitignore-for-a-python-project/
* https://github.com/github/gitignore/blob/master/Python.gitignore

### Fixes

Git trying to commit files that shouldn't be added. Helps prevent accidental mistakes that need to be cleaned up later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,65 @@
-/.project
-/.pydevproject
-*.pyc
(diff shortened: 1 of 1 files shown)


dry923 (approved):
LGTM!

PR #283, opened 2021-06-15
Extend Object-Orientated Structure (Uperf POC)

learnitall:
### Description

Goal of this PR is to submit an initial proof of concept for restructuring snafu to be more object-orientated and pythonic, as well as for replacing argparse with [configargparse](https://pypi.org/project/ConfigArgParse/). Here are the relevant files for review:

* **snafu/config.py**: Module for holding a config manager class which takes care of setting up parsers for benchmarks
* **snafu/process.py**: Module for handling the sampling of processes run through the shell
* **snafu/registry.py**: Holds the global benchmark registry, which auto-detects benchmarks and populates a dict that maps tool-names to their wrappers.
* **snafu/benchmarks**: New directory to hold benchmark modules
    * **snafu/benchmarks/\__init\_.py**: Import relevant objects from snafu/benchmarks/_benchmark.py
    * **snafu/benchmarks/_benchmark.py**: Holds ``Benchmark`` abstract base class and common ``BenchmarkResult`` class
    * **snafu/benchmarks/uperf.py**: Uperf benchmark re-written to use modified code

Here is some pseudo-code for what ``run_snafu`` would look like with these changes

```python
import configargparse
from snafu import registry
# setup logging, exporters, etc.
parser = configargparse.get_argument_parser()
parser.add_argument("--config-file", is_config_file=True)
parser.add_argument("--metadata")
args = parser.parse_known_args()
# parse key1=value1,key2=value2,...
metadata: Dict[str, str] = {
    key: value for key, value in map(lambda pair: pair.split("="), args.metadata.strip().split(","))
}
if registry.TOOLS.get(args.tool, None) is not None:
    benchmark = registry.TOOLS[args.tool](metadata=args.metadata)
    results = benchmark.run()
    # export results
else:
    logger.critical(f"Could not find given tool name {args.tool}")
    return
```

And here is what an example config file would look like:

```yaml
workload: test.xml
sample: 3
resourcetype: vm
uuid: 1337-my-uuid
```

### Fixes

* Implements functionality that is used commonly within benchmarks
* Provides an easy method for documenting data exported by each benchmark, through the ``BenchmarkResult`` class and ``TypedDict``s
* Allow for configuration to be loaded from CLI, env or a config file without any extra logic in benchmarks
* Auto-detect benchmarks as they are added to the code-base, no need to continually update a module that maps tool names to wrappers. Additionally, only benchmarks that have all of their dependencies installed will be added into the global too registry, allowing for auto-detection of supported benchmarks
* Provide simple method for adding additional metadata into a benchmark result.
* Provide clear and simple structure for Benchmarks: just subclass ``Benchmark``, add ``tool_name`` and ``args``, write ``setup`` and ``run`` methods using the provided attributes setup in the parent class, profit.


snafu/benchmarks/uperf.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        Run uperf benchmark ``self.config.sample`` number of times.
+
+        Returns immediately if a sample fails. Will attempt to Uperf run three times for each sample.
+        """
+
+        self.logger.info("Running setup tasks.")
+        if not self.setup():
+            self.logger.critical(f"Something went wrong during setup, refusing to run.")
+            return
```
learnitall, line 215: This code would have to be run in every ``run`` method. Maybe remove from the run method, and create a new method called ``get_results`` or something that runs ``setup`` and then ``run``.

snafu/config.py
```diff
@@ -0,0 +1,58 @@
...
+
+class Config:
+    """
+    Class for managing config parameters.
+
+    Uses argparse and configargparse in the background
+    """
+
+    def __init__(self, tool_name: str):
+        self.config: argparse.Namespace = argparse.Namespace()
```
learnitall, line 37: Maybe a better name for this attribute? So its not ``config.config``?

snafu/registry.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Implement a global tool registry that automatically maps tool names to their classes.
+
+This works by having a metaclass (added onto ```wrapper.Wrapper``), add tool classes to into the
+registry dict when the class is created. For more information on metaclasses, see:
+https://www.geeksforgeeks.org/python-metaclasses/.
```
learnitall, line 8: Should remove this url from the docstring.

snafu/benchmarks/_benchmark.py
```diff
@@ -0,0 +1,64 @@
...
+class BenchmarkResult:
+    name: str
+    metadata: Dict[str, Any]
+    config: Dict[str, Any]
+    data: Dict[str, Any]
+    label: str
+
+    def to_json(self) -> str:
+        result: Dict[str, Any] = dict()
+        result.update(self.config)
```
learnitall, line 23: A lot of benchmarks will add their config mappings on the top-level JSON output next to the result data, rather that placing them in a separate ``config`` key, which is why I formatted the JSON result like this. But if we want to change the structure of the benchmark result to something different then here is where we can do that without having to modify each benchmark.

snafu/process.py
```diff
@@ -0,0 +1,77 @@
...
+import datetime
+import logging
+import subprocess
+
+
+class ProcessRun(TypedDict, total=False):
+    rc: int
+    stdout: str
+    stderr: str
+    runtime: float
```
learnitall, line 14: This might be misleading. Meant to represent the time it took to run the process in seconds.

snafu/registry.py
```diff
@@ -0,0 +1,44 @@
...
+
+After snafu loads up and all the classes are created, the registry dict can be accessed at
+``registry.TOOLS``. Key names will be tool names, values will be their wrapper classes.
+"""
+from typing import Dict
+from abc import ABCMeta
+
+
+TOOLS: Dict[str, object] = dict()
```
learnitall, line 17: This type should be changed to ``Dict[str, Benchmark]``.

snafu/registry.py
```diff
@@ -0,0 +1,44 @@
...
+    ...
+    >>> snafu.registry.TOOLS["my_awesome_tool"]().tool_name
+    'my_awesome_tool'
+    """
+
+    def __new__(cls, clsname, superclasses, attributedict):
+        """Called when a new class is created."""
+
+        new_class = super().__new__(cls, clsname, superclasses, attributedict)
+        TOOLS[attributedict["tool_name"]] = new_class
```
learnitall, line 42: If ``tool_name`` isn't set, then the ``KeyError`` this line would raise may not be specific enough. Might want to consider raising an error with a more specific message like: ``Please define a tool_name attribute for your benchmark.``

snafu/registry.py
```diff
@@ -0,0 +1,44 @@
...
+``registry.TOOLS``. Key names will be tool names, values will be their wrapper classes.
+"""
+from typing import Dict
+from abc import ABCMeta
+
+
+TOOLS: Dict[str, object] = dict()
+
+
+class ToolRegistryMeta(ABCMeta):
```
learnitall, line 20: This class is meant to replace the wrapper_factory by automating benchmark discovery: https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/snafu/utils/wrapper_factory.py

learnitall:
Created a demo script to test out running benchmarks. Here it is:

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
import os
import logging
import configargparse
from snafu.benchmarks import load_benchmarks
from snafu.registry import TOOLS


def setup_logging() -> None:
    logger = logging.getLogger("snafu")
    logger.setLevel(logging.INFO)

    FMT = "%Y-%m-%dT%H:%M:%SZ"

    log_format = "%(asctime)s - %(levelname)s - %(processName)s - %(module)s: %(message)s"
    formatter = logging.Formatter(log_format, datefmt=FMT)

    sh = logging.StreamHandler()
    sh.setLevel(logging.DEBUG)
    sh.setFormatter(formatter)
    logger.addHandler(sh)


def setup_parser() -> None:
    parser = configargparse.get_argument_parser(
        description="run a workload",
        add_config_file_help=True,
        add_env_var_help=True,
        default_config_files=["./snafu.yml"],
        ignore_unknown_config_file_keys=False
    )
    parser.add_argument("tool", help="Benchmark to run")
    parser.add_argument("-f", "--config", help="Config file to load", is_config_file=True)
    parser.add_argument("-l", "--labels", help="Metadata to add in results exported by benchmark.")


def main() -> None:
    setup_parser()
    parser = configargparse.get_argument_parser()
    args, unknowns = parser.parse_known_args()
    print(args)
    print(unknowns)

    load_benchmarks()
    tool = TOOLS.get(args.tool, None)
    if tool is None:
        raise ValueError(f"Given tool {args.tool} can't be found!")
        exit(1)
    else:
        setup_logging()
        logger = logging.getLogger("snafu")
        if args.labels:
            metadata = dict(map(lambda pair: tuple(pair.split("=")), args.labels.strip().split(",")))
        else:
            metadata = None
        logger.info(f"Got metadata: {metadata}")
        benchmark_wrapper = tool(metadata=metadata)
        print(parser.format_help())
        print(parser.format_values())
        for result in benchmark_wrapper.run():
            print(result)


if __name__ == "__main__":
    main()
```

Allows for specifying configuration parameters through a config file with the ``-f`` option, with ``./snafu.yml`` being loaded by default if it exists. Can also specify metadata labels using the format: ``--labels=key=value,key2=value2``. Also note that ``configargparse`` will load parameters in the following order: config < env < cli.

snafu/benchmarks/_load_benchmarks.py
```diff
@@ -0,0 +1,49 @@
...
+# -*- coding: utf-8 -*-
+"""Holds function for automagically importing all benchmark modules in ``snafu.benchmarks``"""
+import os
+import pkgutil
+import importlib
+import logging
+from typing import Tuple, List
+
+
+def load_benchmarks(logger: logging.Logger = None) -> Tuple[List[str], List[str]]:
```
whitleykeith, line 11: I feel like this code should be part of the `benchmarks` package's `__init__.py` and be done implicity when built. That way you don't have to manually call this function further upstream

snafu/benchmarks/uperf.py
whitleykeith: You could move the logic in `setup` into the Class `__init__` method

snafu/benchmarks/uperf.py
```diff
@@ -0,0 +1,242 @@
...
+
+
+class UperfConfig(TypedDict, total=False):
+    # Parsed from stdout
+    test_type: str
+    protocol: str
+    message_size: int
+    read_message_size: int
+    num_threads: int
+    duration: int
```
whitleykeith, line 25: I think these are results and not configurations. The current implementation of uperf simply adds this data to the individual datapoints in it's `_json_payload()` so I would consider this part of `RawUperfResult`

snafu/benchmarks/uperf.py
```diff
@@ -0,0 +1,242 @@
+#!/usr/bin/env python3
```
whitleykeith, line 1: i think we probably want to have a directory called `uperf` here instead of putting the file directly under `benchmarks`. Each benchmark will probably have it's own tests, dockerfile, and potentially config file.

Something like

benchmarks/
   uperf/
      wrapper.py
      config.yaml
      Dockerfile

rsevilla87: I was thinking the the same, we can also consider creating subpackages within the benchmarks module

jtaleric: agreed... We used to house all the other non-code stuff outside of `snafu/benchmark-wrapper` but that made the build of the container images more complex... I prefer we keep it together, but I am open to ideas if we want to break it out again...

snafu/config.py
jtaleric: `config.params` ?

jtaleric (commented):
Impressive @learnitall!

jtaleric:
What CI is choking on 
```
15:53:40 Traceback (most recent call last):
15:53:40   File "/usr/local/bin/run_snafu", line 33, in <module>
15:53:40     sys.exit(load_entry_point('snafu', 'console_scripts', 'run_snafu')())
15:53:40   File "/usr/local/bin/run_snafu", line 25, in importlib_load_entry_point
15:53:40     return next(matches).load()
15:53:40   File "/usr/local/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 167, in load
15:53:40     module = import_module(match.group('module'))
15:53:40   File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
15:53:40     return _bootstrap._gcd_import(name[level:], package, level)
15:53:40   File "<frozen importlib._bootstrap>", line 994, in _gcd_import
15:53:40   File "<frozen importlib._bootstrap>", line 971, in _find_and_load
15:53:40   File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
15:53:40   File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
15:53:40   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
15:53:40   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
15:53:40   File "/opt/snafu/snafu/run_snafu.py", line 29, in <module>
15:53:40     from snafu.benchmarks import load_benchmarks
15:53:40   File "/opt/snafu/snafu/benchmarks/__init__.py", line 5, in <module>
15:53:40     from snafu.benchmarks._benchmark import Benchmark, BenchmarkResult
15:53:40   File "/opt/snafu/snafu/benchmarks/_benchmark.py", line 6, in <module>
15:53:40     from dataclasses import dataclass
15:53:40 ModuleNotFoundError: No module named 'dataclasses'```

**Comment by mfleader** [comment]
> What CI is choking on
> 
> ```
> 15:53:40 Traceback (most recent call last):
> 15:53:40   File "/usr/local/bin/run_snafu", line 33, in <module>
> 15:53:40     sys.exit(load_entry_point('snafu', 'console_scripts', 'run_snafu')())
> 15:53:40   File "/usr/local/bin/run_snafu", line 25, in importlib_load_entry_point
> 15:53:40     return next(matches).load()
> 15:53:40   File "/usr/local/lib/python3.6/site-packages/importlib_metadata/__init__.py", line 167, in load
> 15:53:40     module = import_module(match.group('module'))
> 15:53:40   File "/usr/lib64/python3.6/importlib/__init__.py", line 126, in import_module
> 15:53:40     return _bootstrap._gcd_import(name[level:], package, level)
> 15:53:40   File "<frozen importlib._bootstrap>", line 994, in _gcd_import
> 15:53:40   File "<frozen importlib._bootstrap>", line 971, in _find_and_load
> 15:53:40   File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
> 15:53:40   File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
> 15:53:40   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
> 15:53:40   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
> 15:53:40   File "/opt/snafu/snafu/run_snafu.py", line 29, in <module>
> 15:53:40     from snafu.benchmarks import load_benchmarks
> 15:53:40   File "/opt/snafu/snafu/benchmarks/__init__.py", line 5, in <module>
> 15:53:40     from snafu.benchmarks._benchmark import Benchmark, BenchmarkResult
> 15:53:40   File "/opt/snafu/snafu/benchmarks/_benchmark.py", line 6, in <module>
> 15:53:40     from dataclasses import dataclass
> 15:53:40 ModuleNotFoundError: No module named 'dataclasses'```
> ```

This PR requires python ^3.7 because of `dataclasses`. What is the python version you're using?

**Review by mfleader — CHANGES_REQUESTED** [review]
vermin output for the `benchmarks` directory

```shell
vermin --violations --target=3.6 benchmarks

~2, ~3       /home/mleader/workspace/cloud-bulldozer/benchmark-wrapper/snafu/benchmarks/__init__.py
!2, 3.7      /home/mleader/workspace/cloud-bulldozer/benchmark-wrapper/snafu/benchmarks/_benchmark.py
  'dataclasses' module requires !2, 3.7

!2, 3.6      /home/mleader/workspace/cloud-bulldozer/benchmark-wrapper/snafu/benchmarks/_load_benchmarks.py
!2, 3.8      /home/mleader/workspace/cloud-bulldozer/benchmark-wrapper/snafu/benchmarks/uperf.py
  'typing.TypedDict' member requires !2, 3.8

Tip: You're using potentially backported modules: dataclasses, importlib, typing
If so, try using the following for better results: --backport dataclasses --backport importlib --backport typing

Minimum required versions: 3.8
Incompatible versions:     2
Target versions not met:   3.6
```

snafu/benchmarks/_load_benchmarks.py
learnitall: That's super smart, I like that. I think I might keep the code separate, just to keep stuff organized, but I'll add an implicit call.

learnitall:
Currently there is a problem with dependencies that we need to look into. For some of these changes that involve typing and new data structures, we need upstream ports. Might be work looking at doing some dependency work before merging these changes in.

snafu/benchmarks/uperf.py
learnitall: I popped uperf back into a directory, using the original structure. I originally made the change, as I personally am a fan of keeping code and non-code separate. Maybe we can discuss this offline some more in the future and revisit our build process.

snafu/benchmarks/uperf.py
learnitall: Got it OK. In ``_json_payload``, these data are labeled as ``config`` (see https://github.com/cloud-bulldozer/benchmark-wrapper/blob/c649d9db9c0dc3dd563b698cad759d74209b2d79/snafu/uperf_wrapper/trigger_uperf.py#L107) which is why I did the same. Didn't realize it was part of the result, I'll make a change.

learnitall: Hmm, looking at benchmark-operator, the variables in question are set as arguments for the benchmark. Check out: https://github.com/cloud-bulldozer/benchmark-operator/blob/master/roles/uperf/templates/workload.yml.j2#L140-L152

I think that there are a couple of different terms that may be getting confused here. Here is how I see it:

* ``UperfConfig``: contains all the variables that were given as arguments to setup the benchmark
* ``RawUperfResult``: contains a statistic that was emitted from the benchmark
* ``UperfStdout``: contains stuff that is parsed from the output of Uperf
* ``UperfResultData``: contains the parsed stdout statistics

I think that instead of adding the variables in question to ``RawUperfResult``, I'll change the names of ``RawUperfResult`` and ``UperfResultData`` to ``RawUperfStat`` and ``UperfStat`` respectively. Thoughts?

snafu/config.py
learnitall: Done deal.

snafu/registry.py
learnitall: Can't set to ``Dict[str, Benchmark]``, as doing so would require importing the Benchmark class, creating a circular import. Leaving as-is.

learnitall:
On next success of a CI test, will be ready to merge IMO.

snafu/benchmarks/_benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+
+            labels[key] = value
+        return labels
+
+
+class Benchmark(ABC, metaclass=registry.ToolRegistryMeta):
+    """
+    Abstract Base class for benchmark tools.
+
+    To use, subclass, set the ``tool_name`` and ``args`` attribute, and overwrite the ``run``, ``cleanup`` and
```
learnitall, line 53: Outdated comment, need to update

learnitall:
Turns out that the Uperf CI tests weren't running, due to the new folder structure. See Commit 
04b50d2. I'm testing the benchmark now to fix any issues locally, then I'll rerun to verify all is well.

whitleykeith (approved):
> 95 commits

[lgtm!](https://youtu.be/rR4n-0KYeKQ)

snafu/benchmarks/uperf/uperf.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+class ParseRangeAction(FuncAction):
+    """Parses node_range and density_range attributes."""
+
+    def func(self, arg: str) -> List[int]:
+        return [int(x) for x in arg.split("-") if x != ""]
+
+
+@dataclasses.dataclass
```
kurtismullins, line 21: I'm glad to see you're taking advantage of the newer `dataclass` type. Should probably mention the Python version requirements in the docs unless I overlooked it.

kurtismullins (commented):
I'm looking through your code as I'm going to make some changes to Vegeta and planned on adopting your OOP idea.

I do have concerns about handling the data -- specifically, I don't want to write potentially hundreds of megabytes, if not gigs, to memory just to push them off to ES. Currently, I can lazy evaluate the files already generated by vegeta and use ES' `bulk_write()` functionality to lazy read+write them in batches which uses very low memory (perfect for running tests in pods) and also extremely quick (compared to generating massive HTTP requests in memory).

snafu/benchmarks/uperf/uperf.py
learnitall: I can add a note in the docs about it for sure, but I did add the dataclasses backport into our dependencies, meaning our version requirements haven't changed.

snafu/benchmarks/_benchmark.py
```diff
@@ -0,0 +1,129 @@
...
+from snafu import registry
+from snafu.config import Config, ConfigArgument, FuncAction
+
+
+@dataclass
+class BenchmarkResult:
+    name: str
+    metadata: Dict[str, Any]
+    config: Dict[str, Any]
+    data: Dict[str, Any]
```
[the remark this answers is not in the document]
learnitall: I might be confused on what you're asking here. The BenchmarkResult is an in-memory container for a single result or datapoint from the benchmark, not for the entire run. All it does is replace the dictionaries that benchmarks were yielding so we can start to transition to a more standard data model across benchmarks and provide some readability. The Benchmark.collect method still uses the generator interface to yield these results when passing to py_es_bulk

snafu/benchmarks/_load_benchmarks.py
```diff
@@ -0,0 +1,70 @@
...
+        if show_tb and len(self.errors) > 0:
+            logger.log(level, f"Got the following errors:")
+            for benchmark, exc_info in self.errors.items():
+                tb = "".join(traceback.format_exception(*exc_info))
+                logger.log(level, f"Benchmark module {benchmark} failed to import:\n{tb}")
+
+
+def load_benchmarks() -> DetectedBenchmarks:
+    """
+    Autodetect modules in same directory as source file (``__file__``) and automatically import them.
```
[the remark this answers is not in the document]
learnitall: This replaces the static method of benchmark discovery that we currently have in ``utils/wrapper_factory.py``, which keeps a mapping of benchmark names to wrapper objects. My thinking was that this would reduce maintenance as we don't have to keep this file up-to-date as benchmarks are added and removed. All a user has to do is place their module under ``benchmarks``, no other code has to be modified. Additionally, this dynamic import system means that we don't have to ship snafu with dependencies for each benchmark. The current wrapper factory requires importing each benchmark, which means each benchmark's python dependencies have to be present, otherwise an import error is raised. This new method will catch these import errors, log them, and only expose modules that were successfully imported.

snafu/benchmarks/_benchmark.py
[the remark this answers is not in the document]
kurtismullins: Ahh okay. I'm going to dig in deeper and make sure I understand how it works. I haven't worked on Snafu's code in the past so I may have misunderstood the whole flow of data. 

Thanks for pointing that out!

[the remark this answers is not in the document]
learnitall: For sure! Thanks for asking about it. I'd recommend taking a look at ``run_snafu.py``, as that's where the data flow can be seen.

docs/source/install.md
```diff
@@ -1 +1,30 @@
 # Installation
+
+Installing benchmark-wrapper is all done through pip and git, requiring Python >= 3.6. For instance, to download benchmark-wrapper and install within a new virtual environment:
```
mfleader, line 3: >= 3.7?

setup.cfg
```diff
@@ -21,14 +21,14 @@ zip_safe = False
 packages = find:
 include_package_data = True
 # Add here dependencies of your project (semicolon/line-separated), e.g.
-install_requires = configparser; elasticsearch>=6.0.0,<=7.0.2; statistics; numpy; pyyaml; requests; redis; python-dateutil>=2.7.3; prometheus_api_client; scipy; openshift==0.11; kubernetes==11; setuptools>=40.3.0; boto3; flent; importlib_metadata; kafka-python; ttp
+install_requires = dataclasses; configargparse; configparser; elasticsearch>=6.0.0,<=7.0.2; statistics; numpy; pyyaml; requests; redis; python-dateutil>=2.7.3; prometheus_api_client; scipy; openshift==0.11; kubernetes==11; setuptools>=40.3.0; boto3; flent; importlib_metadata; kafka-python; ttp
 # tests_require = pytest; pytest-cov
 # Require a specific Python version, e.g. Python 2.7 or >= 3.4
 python_requires = >=3.6
```
mfleader, line 27: dataclass module requires 3.7

snafu/registry.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+"""
+Implement a global tool registry that automatically maps tool names to their classes.
+
+This works by having a metaclass (added onto ```wrapper.Wrapper``), add tool classes to into the
```
mfleader, line 6: change to "add tool classes **to** the registry"

snafu/config.py
```diff
@@ -0,0 +1,206 @@
...
+    Returns
+    -------
+    bool
+        ``True`` if the given file exists and has the given perms, else ``False``
+    """
+
+    if perms is None:
+        perms = os.R_OK
+    perms |= os.F_OK
+    return os.access(os.path.abspath(file), perms)
```
mfleader, line 37: I want to suggest pathlib here, but I don't think this is ever going to get used on Windows.

setup.cfg
learnitall: The dataclass module is shipped with 3.7, but this is the 3.6 backport. See: https://pypi.org/project/dataclasses/

snafu/config.py
learnitall: For sure, I would be surprised if we did use this on Windows. I'm not familiar with how this function would behave on a Window's system, nor with pathlib, but it seems like we might be OK for now. Pathlib seems to be a super nice way to do these operations, but because using os does work ok here, I'm going to resolve this for now.

learnitall: One thing we might want to think about, for future reference, is that ``os.access`` is pretty lazy when it comes to determining permissions, i.e., it doesn't check SELinux permissions: https://bugs.python.org/issue42481

snafu/benchmarks/uperf/uperf.py
```diff
@@ -0,0 +1,301 @@
...
+    iteration: Optional[int] = None
+
+
+class Uperf(Benchmark):
+    """
+    Wrapper for the uperf benchmark.
+    """
+
+    tool_name = "uperf"
+    args = (
```
rsevilla87, line 93: Hey @learnitall, small nit. Why do we need to define these flags as ConfigArgument?

docs/source/contributing/adding_workloads.ipynb
```diff
@@ -0,0 +1,1165 @@
+{
```
rsevilla87, line 1: Curious, why are we adding this docs in Jupyter notebooks format?

rsevilla87 (approved):
LGTM!, keep it up!

docs/source/contributing/adding_workloads.ipynb
learnitall: Main idea was that the notebook could be downloaded and used by users interactively as an example. Plus it renders pretty nice within sphinx and makes adding in large code blocks that should be executable fairly easy.

learnitall: Additionally I wanted to test it out- it was pretty fun!

learnitall: Here is what the rendered page looks like: https://my-benchmark-wrapper-fork.readthedocs.io/en/latest/contributing/adding_workloads.html. Adding in all of the code, and the code's output was super easy to do and test within a notebook.

PR #293, opened 2021-07-09
enable prometheus data collection for uperf workload

mukrishn:
### Description
This change will let user to enable prometheus metric collection and forward to elasticsearch for uperf workoads, just only collects CPU metrics at present.

### DIFF
--- a/snafu/uperf_wrapper/trigger_uperf.py
+++ b/snafu/uperf_wrapper/trigger_uperf.py
@@ -44,6 +44,7 @@ class Trigger_uperf:
         self.density_range = args.density_range
         self.node_range = args.node_range
         self.pod_id = args.pod_id
+        self.prom_es = args.prom_es
 
     def _json_payload(self, results, data, sample):
         processed = []
@@ -120,6 +121,10 @@ class Trigger_uperf:
                 "read_message_size": int(rsize),
                 "num_threads": int(nthr),
                 "duration": len(results),
+                "tool": "uperf",
+                "test_config": "uperf",
+                "starttime": float(results[0][0]) / 1000,
+                "endtime": float(results[-1][0]) / 1000,
             },
         )
 
@@ -144,7 +149,10 @@ class Trigger_uperf:
             documents = self._json_payload(results, data, s)
             if len(documents) > 0:
                 for document in documents:
-                    yield document, "results"
+                    if self.prom_es == "":
+                        yield document, "results"
+                    else:
+                        yield document, "get_prometheus_trigger"
             logger.info(data)
             logger.info(stdout)
             logger.info("Finished executing sample %d out of %d" % (s, self.sample))
--- a/snafu/uperf_wrapper/uperf_wrapper.py
+++ b/snafu/uperf_wrapper/uperf_wrapper.py
@@ -60,6 +60,7 @@ class uperf_wrapper:
         # each node will run with density number of pods, this is the 0 based
         # number of that pod, useful for displaying throughput of each density
         self.args.pod_id = os.getenv("my_pod_idx", "")
+        self.args.prom_es = os.getenv("prom_es", "")
 
     def run(self):
         uperf_wrapper_obj = Trigger_uperf(self.args)
--- a/snafu/utils/prometheus_labels/uperf_included_labels.json
+++ b/snafu/utils/prometheus_labels/uperf_included_labels.json
@@ -1,0 +1,8 @@
+{
+    "data": {
+        "Average_CPU_Usage_per_Instance": {
+            "label": "node_cpu_seconds_total",
+            "query": "(sum by (instance, mode)(rate(node_cpu_seconds_total{mode='idle'}[1m])))"
+        }
+    }
+}


jtaleric (commented):
so, I spoke with @acalhounRH and @smalleni about this PR. 

Here is the take away(s) - 

- Using Ripsaw/Benchmark-Operator we will ONLY get the start to end of the metrics.

The issue with this, UPerf iterates through the entire test scenarios defined in the XML. We won't have a way to pick granular data, such as TCP/16384/2threads. 

- Prometheus in benchmark-wrapper / snafu

This builds on the concept of a collector.

Proposal(s) - 
- Either we spend the time to harden the collector concept, and build upon the original work Alex built. 
- Enhance the collection from Ripsaw. We have already discussed passing more detailed state from SNAFU -> Ripsaw (or higher level orchestration). We could provide more detailed state back to a higher-level tool via Redis. Which would allow Ripsaw to start/stop collection and annotate the state with the kube-burner collection/indexing. 

Thoughts?

learnitall:
I personally am a fan of starting with having snafu export state, which we can then have Ripsaw pick up on. We've already talked internally a bit on the collector concept and I feel the advantages of keeping snafu and collectors separate are worth sticking to. If we allow for the state information to be detailed enough, we could enable external tools to collect metrics on the level of benchmark runs and iterations.

I propose we create a structure within snafu for capturing and exporting state, with the ability to add custom state export locations as needed. If the exported state includes stage and timestamp, we should be able to pull metrics from prom during the target time intervals, similar to how the crucible project works.

rsevilla87:
> so, I spoke with @acalhounRH and @smalleni about this PR.
> 
> Here is the take away(s) -
> 
>     * Using Ripsaw/Benchmark-Operator we will ONLY get the start to end of the metrics.
> 
> 
> The issue with this, UPerf iterates through the entire test scenarios defined in the XML. We won't have a way to pick granular data, such as TCP/16384/2threads.
> 
>     * Prometheus in benchmark-wrapper / snafu
> 
> 
> This builds on the concept of a collector.
> 
> Proposal(s) -
> 
>     * Either we spend the time to harden the collector concept, and build upon the original work Alex built.
> 
>     * Enhance the collection from Ripsaw. We have already discussed passing more detailed state from SNAFU -> Ripsaw (or higher level orchestration). We could provide more detailed state back to a higher-level tool via Redis. Which would allow Ripsaw to start/stop collection and annotate the state with the kube-burner collection/indexing.
> 
> 
> Thoughts?

In my opinion to perform tasks such as system metrics collection we should favor solutions built in higher layers such as benchmark-operator. As we've discussed several times our ultimate goal with this project (snafu/benchmark-wrapper) is to provide the logic required to execute, parse, process and index results from different benchmark tools.

Snafu/benchmark-wrapper is a tool that should be able to run out of k8s environments, and prometheus is a monitoring system tightly coupled to these environments, we should avoid increasing the dependency of running snafu on top of k8s.

I think  a good solution would be creating a sort of signaling mechanism capable of providing a status of the benchmark execution, so higher layers can be aware when each iteration of the test matrix happened. This way we can improve the current system-metrics collection feature and make it capable of using this information to collect metrics using the start/end dates of each test case.

learnitall:
Please see https://github.com/cloud-bulldozer/benchmark-wrapper/issues/295

smalleni:
> > so, I spoke with @acalhounRH and @smalleni about this PR.
> > Here is the take away(s) -
> > ```
> > * Using Ripsaw/Benchmark-Operator we will ONLY get the start to end of the metrics.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The issue with this, UPerf iterates through the entire test scenarios defined in the XML. We won't have a way to pick granular data, such as TCP/16384/2threads.
> > ```
> > * Prometheus in benchmark-wrapper / snafu
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This builds on the concept of a collector.
> > Proposal(s) -
> > ```
> > * Either we spend the time to harden the collector concept, and build upon the original work Alex built.
> > 
> > * Enhance the collection from Ripsaw. We have already discussed passing more detailed state from SNAFU -> Ripsaw (or higher level orchestration). We could provide more detailed state back to a higher-level tool via Redis. Which would allow Ripsaw to start/stop collection and annotate the state with the kube-burner collection/indexing.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Thoughts?
> 
> In my opinion to perform tasks such as system metrics collection we should favor solutions built on higher layers such as benchmark-operator. As we've discussed several times our ultimate goal with this project (snafu/benchmark-wrapper) is to provide the logic required to execute, parse, process and index results from different benchmark tools.
> 
> Snafu/benchmark-wrapper is a tool that should be able to run out of k8s environments, and prometheus is a monitoring system tightly coupled to these environments, we should avoid increasing the dependency of running snafu on top of k8s.
> 
> I think a good solution would be creating a sort of signaling mechanism capable of providing a status of the benchmark execution, so higher layers can be aware when each iteration of the test matrix happened. This way we can improve the current system-metrics collection feature and make it capable of using this information to collect metrics using the start/end dates of each test case.

I have to agree with Raul here, exporting state from SNAFU and having that picked up the orchestrator would be the best path or else we risk going against design goals for the wrapper.

bengland2:
@acalhounRH cc'ed.  As I recall,  I think Alex was trying to create an environment where you could look at prom metrics for a particular kind of workload that was going on.   For example, with fio, if it's running 1,4,16,64,256,1024,4096-KiB rand{read,write} and you just want your dashboard to show you what CPU utilization was for rand writes, you can do that if the Documents submitted by benchmark-wrapper include the test type.   And in fact I think there is at least one version of the grafana dashboard for fio that did this at one point in time.   You can't do that without the test type being embedded in the document.  Otherwise you have to know exactly when the test type you are interested in starts and stops, and look for it on the time X-axis.   This is a whole lot harder to do (accurately), and most people won't bother.    As long as there is some way to include test type in the documents, then I'd be open to other ways of implementing prom collection that might be more benchmark-independent.

jtaleric:
> * Enhance the collection from Ripsaw. We have already discussed passing more detailed state from SNAFU -> Ripsaw (or higher level orchestration). We could provide more detailed state back to a higher-level tool via Redis. Which would allow Ripsaw to start/stop collection and annotate the state with the kube-burner collection/indexing.

Out of the two methods I mentioned, it seems that this is the favorite.. Which we have a PR to pass state....

jtaleric:
https://coreos.slack.com/archives/GTHU29259/p1628013827128700 

Message from @mukrishn that we will close this out for now.

PR #299, opened 2021-07-16
Create Staged CI Pipeline

learnitall:
### Description

This is a big one, so strap in. This PR adds a bunch of changes in order to test the quality and functionality of our code. I'll walk through the main changes one by one:

#### New pre-commit hooks for code quality and bug catching:

The goal with adding these additional pre-commit hooks is to add checks which will help keep our code clean and readable. I've been developing with these hooks setup for a couple of weeks now and not only have they have caught dozens of errors in my code, they've helped keep my code readable.

  * [isort](https://pypi.org/project/isort/): works in the same way as black to automatically sort imports according to PEP standards
  * [pylint](https://pylint.org/): has checks which flake8 doesn't include; they are a good pair
  * [pydocstyle](http://www.pydocstyle.org/en/stable/): used for docstring enforcement.
  * [mypy](http://mypy-lang.org/): ensures that type hints all agree with each other
  * [pyupgrade](https://pypi.org/project/pyupgrade/): checks for older syntax expressions that can be refactored to take advantage of newer python syntax
 
I've disabled a lot of the more intrusive checks, but these hooks will break files and may need to be skipped until we have time to refactor our code. For instance, if I want to make a small change in a wrapper as a user, pre-commit may force me to fix a bunch of linting errors before being able to commit the change. Hooks can easily be skipped like so: `SKIP=pylint git commit`.

I understand these are intrusive, but I wanted to propose them because I feel in the long run they will be worth the extra effort. I think a nice compromise would be to not include the new hooks for now, but have our transitioned benchmark code be up-to-date with the new hooks' requirements.

#### Tox/Pytest Testing Structure

I created a structure to allow for unit and functional testing. It takes advantage of [pytest](https://docs.pytest.org/en/6.2.x/) and [tox](https://tox.readthedocs.io/en/latest/config.html). Unit tests are located within the ``tests/unit`` directory while functional tests are located within ``tests/functional``. I've configured tox with the following environments:

* Unit tests: ``py36``, ``py37``, ``py38``, ``py39``
* Documentation tests: ``docs``
* Functional tests: ``functional``

I've also configured unit tests to generate coverage reports that can be sent off to [codecov.com](https://about.codecov.io/) which allows us to visual which sections of our code are covered by our unit tests and which sections need some love.

Functional tests were a bit tricky to implement, but woah is the result cool. Under ``tests/functional`` there is the following directory structure:

* ``tests/functional/benchmark``: Folder containing functional test resources for a benchmark
* ``tests/functional/benchmark/deploy.yaml``: Kubernetes manifest which describes a pod to deploy before a test suite is run
* ``tests/functional/benchmark/test_name.py``: Module containing the actual tests to run

I've written a new fixture called ``manifest`` which is scoped at a module level. Whenever a test in a Python module uses this fixture, the fixture will find the companion ``deploy.yaml`` file that exists in the same directory as test module, deploy the manifest using rootless [podman play kube](http://docs.podman.io/en/latest/markdown/podman-play-kube.1.html), run all of the tests within the module, then tear-down any pods that were deployed with the label ``test=snafu``. Unfortunately podman play doesn't have a teardown feature yet, so we have to delete pods manually by label.

An example of this structure can be seen in the Uperf tests that I've written for this PR.

What's cool about this is that we can use different pods for different tests, all for the same benchmark. And pods can be reused for multiple tests. For instance, if I wanted to have two different versions of elasticsearch while testing FIO, I could use the following directory structure:

* ``tests/functional/fio/es6/{deploy.yaml,test_fio_es6.py}``
* ``tests/functional/fio/es7/{deploy.yaml,test_fio_es7.py}``

All of the tests in ``test_fio_es6.py`` will use the same pod defined by ``es6/deploy.yaml`` and all of the tests within ``test_fio_es7.py`` will use the same pod defined by ``es7/deploy.yaml``

Additionally, I've added two new command line arguments to pytest which control how pods are created and cleaned. Using ``--keep-pods`` will tell pytest to not delete pods after the test is complete, while using ``--use-existing-pods`` will tell pytest that the pod(s) for a test are already deployed and don't need to be setup. This means that I can run a test using ``tox -e functional -- --keep-pods`` and if the test fails, nothing is cleaned up, allowing for easy debug.

Finally, I've configured pytest to automatically mark functional tests with the appropriate benchmark name, which means that we can run all tests for ONLY a certain benchmark with the following: ``tox -e functional -- -k benchmark_name``. The benchmark name is determined from the directory which follows ``tests/functional``, so all tests under ``tests/functional/uperf`` will be marked with ``uperf``.

This change allows us to test fully test snafu on our workstations, using just Python and rootless podman.

#### GitHub Actions CI

Lastly, I've created new workflow files which will run our linting, unit and functional tests within GitHub actions. The workflows are configured to run on a push or a pull_request and can be triggered manually. You can see example runs over at my fork: https://github.com/learnitall/benchmark-wrapper/runs/3089605349?check_suite_focus=true

The first workflow runs linting tests, an install test, unit tests and doc tests for Python 3.6-3.9. It's pretty much the same as the ``ci.yaml`` workflow that we had before, but with added pytest stuffs. I've configured this workflow to upload coverage reports to Codecov, so if we want to keep this feature we will need to setup codecov for snafu under cloud-bulldozer.

The second workflow (functional.yaml) will do the following:

* Autodetect all Dockerfiles and Containerfiles within the repository and determine the associated benchmark (based on where it was found) and architecture (based on the extension). 
* Determine which images need to be built and tested based off of what files were changed
* In parallel, for each of the changed images:
  * Build each of the container files
  * Push to quay.io, tagged with the GitHub commit SHA. Auth is configured through GitHub secrets
  * Run the associated functional tests

The auto-detect functionality is provided through a bash script located at ``ci/discover_cfs.sh``, which outputs parsable-JSON to configure a [GH actions matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix)

The cool part about this is that:
1. We only run the tests for benchmarks that have been changed.
2. If a change impacts all benchmarks, then all benchmarks will be tested.
3. Built images are pushed to quay so we can perform further analysis as needed. To see an example of this, my quay account is [rydrew](https://quay.io/user/rydrew)
4. Images pushed to quay are identified by commit, therefore there is no potential for overlap between commits
5. We can manually trigger builds using the GH Actions UI.
6. Workflows can be run in forks, meaning PR can be thoroughly tested before they are submitted.

One awesome note is that I've configured the manual trigger to allow us to compare any two branches, not just between master. This can help development by allowing us to test differences between any two pieces of our code. In my fork I can create a branch, make some changes, create a branch off of that branch, make some changes, and only test the differences between those two branches. Can really help with the debug process if you're into branching.

Screenshot from 2021-07-16 15-28-59

There are a couple of notes that I have which we need to figure out before we can merge:
* We need to modify the functional.yaml workflow so it works for our quay account and isn't linked to mine. Auth also needs to be setup within our secrets
* The ppc64le builds take longer than an hour. I can move them to a separate workflow if we'd like?
* The following AMD64 images currently have failing builds: linpack, oslat, cyclictest.

Thanks! Sorry for the wall of text!


learnitall:
Looks like GH Actions ran for this PR, cool! I'll look into the errors on Monday.

.github/workflows/code_check.yml
```diff
@@ -0,0 +1,72 @@
...
+  - push
+  - pull_request
+  - workflow_dispatch
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
```
amitsagtani97, line 12: Isn't code quality check using PEP8 standard independent of python version?

.github/workflows/functional.yml
```diff
@@ -0,0 +1,109 @@
...
+    inputs:
+      diff_branch:
+        description: 'Upstream branch for determining changes'
+        default: 'master'
+      run_all:
+        description: 'Run all tests regardless of changes?'
+        default: 'false'
+env:
+  IMAGE_REGISTRY: quay.io
+  IMAGE_USER: rydrew
```
amitsagtani97, line 15: maybe, replace it with redhat_perf_ci account and store credentials in the repository secrets.

learnitall: Yep for sure. There are a couple of changes that need to be made to transition from my test accounts over to the more generic ones we would use for cloud-bulldozer.

.github/workflows/code_check.yml
learnitall: Moved code check for linting over to pre-commit CI. We just need to enable it by giving it proper permissions to the repo.

learnitall:
I didn't like how the each job would appear in the UI- it was super messy having 15+ jobs all thrown in there. I consolidated the unit tests and documentation tests into one job, but kept our ability to run the tests against each Python version. Now:

* Unit tests are ran within a different workflow than documentation tests
* Image build tests will only run if unit tests succeed
* Unit tests and documentation tests will run each suite of tests against Python 3.6, 3.7 and 3.8 sequentially. We lose parallel execution, but it only takes about 5 minutes anyways.

learnitall:
For now I am disabling having the functional tests be triggered by the successful completion of the "Unit Tests" workflow. Workflows can only be triggered by another workflow if they are in the default branch, therefore the functional tests right now will not run for this PR. Changing our setup temporarily so we can test out the functional test CI. After we merge, here is what the top of the functional.yml file needs to be changed to in order to enable the more 'staged' setup:

```yaml
name: Build Images
on:
  workflow_run:
    workflows: ["Unit Tests"]
    types: [completed]
...
jobs:
  find-cfs:
    if: >-
        github.event_name != 'workflow_run' ||
        github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success'
```

learnitall:
I'm separating out the image build step from the image push step, in an effort to prevent "pwn requests", (https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) as pushing images requires access to secrets.

Two items that we may need to figure out is are:
* ppc64le builds all time out. This may be because I have the amd64 runners that GitHub provides emulating ppc64le in order to do the build. We could think about hosting our own GitHub runner that uses the architecture to speed things up or remove them from the CI.
* I'm going to work tomorrow on integrating the 'ok-to-test' label in, as the CI is currently set to rebuild on every push and it is getting spammed like crazy while I'm trying to test.

learnitall:
I've disabled multi-arch builds while we figure out what we want to do with our ppc64le Dockerfiles. No point in running them over and over again if we know they time out.

.github/workflows/push-tests-test.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps:
+      - name: Install podman
+        run: sudo apt update && sudo apt install -y podman
+      - name: Download all built images
+        run: |
+            set -x
+            base_url="https://api.github.com/repos"
+            token_header="authorization: Bearer ${{ secrets.GITHUB_TOKEN }}"
+            repo="${{ github.repository }}"
+            run_id="1056706962"
```
amitsagtani97, line 18: is this a random number?

learnitall:
This PR is super huge. I'm realizing that it needs to be split into different parts in order to make review a bit easier. Closing this out, will open new PRs this week.

PR #303, opened 2021-07-20
so we know which metrics are being pulled

bengland2:
just add a log message saying where the prom metrics are being obtained so that it's easy to see what should be in elasticsearch after the test completes


bengland2:
tested and saw the message in the snafu log, however I hit another bug in the code - if it fails to get any response to its query for a metric, then it blows up and fails to get ANY further metrics, even if they exist, very bad.   Fixing this now...

bengland2:
retested with test_fiod.sh, saw much more prom data and saw log message, merging...

PR #304, opened 2021-07-20
Add dnsperf wrapper

mfleader:
### Description

This module adds a wrapper for [dnsperf](https://www.dns-oarc.net/tools/dnsperf), to execute a sequence of stress tests on a system's DNS service, and report the round trip time in microseconds for each DNS query made in the test. The new  `tool_name` for the dnsperf wrapper is `dnsperf`.

I removed the old `dns_perf_wrapper`. I will squash the commits on merge.

### DIFF
--- a/snafu/utils/get_prometheus_data.py
+++ b/snafu/utils/get_prometheus_data.py
@@ -67,6 +67,7 @@ class get_prometheus_data:
                 filename = tool_include_file
             else:
                 filename = os.path.join(include_file_dir, "included_labels.json")
+            logger.info("using prometheus metric include file %s" % filename)
 
             # open tools include file and loop through all
             with open(filename, "r") as f:
@@ -84,6 +85,8 @@ class get_prometheus_data:
                     response = self.pc.custom_query_range(query, self.start, self.end, step, None)
 
                 except Exception as e:
+                    # response undefined at this point, we want to skip next for loop
+                    response = []
                     logger.info(query)
                     logger.warn("failure to get metric results %s" % e)
 


snafu/benchmarks/dnsperf/dnsperf.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def __str__(self) -> str:
+        return (
+            f"Name: {self.fqdn}, "
+            f"Qtype: {self.qtype}, "
+            f"Rcode: {self.rcode}, "
+            f"Rtt microseconds: {self.rtt_mu_s}"
+        )
+
+    def __repr__(self) -> str:
```
learnitall, line 35: Was the goal of adding the ``__repr__`` method here to ensure that the dataclass had a repr method, or was it to override the one that dataclasses autogenerate?

snafu/benchmarks/dnsperf/Dockerfile
```diff
@@ -0,0 +1,44 @@
+FROM registry.centos.org/centos:8
```
learnitall, line 1: CentOS 8 hits its EOL in December. I know you're using the stream repo below, but I think it might be a good idea to use a different base image as support is kind of in the air.

snafu/benchmarks/dnsperf/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    ln -s /usr/local/bin/python3.7 /usr/local/bin/python && \
+    python -m pip install -U pip && \
+    mkdir --parent /opt/snafu
+
+COPY . /opt/snafu
+
+RUN python -m pip install -e /opt/snafu
+
+# dnsperf dependencies
+RUN dnf install -y --nodocs epel-release && \
```
learnitall, line 36: Can we move this RUN command to be before the snafu installation? This way during local development/testing if we change something to snafu we don't have to also reinstall dnsperf: https://gabnotes.org/docker-images-layers-and-cache/. Maybe move all of these dnf steps into the first RUN command?

snafu/benchmarks/dnsperf/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        gcc \
+        libffi-devel \
+        make \
+        openssl-devel \
+        xz-devel \
+        zlib-devel && \
+    curl https://www.python.org/ftp/python/3.7.7/Python-3.7.7.tgz -o /opt/Python-3.7.7.tgz && \
+    tar xzf /opt/Python-3.7.7.tgz --directory /opt
+
+RUN dnf install -y \
```
learnitall, line 18: Why are we installing redis and the ocp client? Is that for benchmark-operator?

snafu/benchmarks/dnsperf/dnsperf.py
```diff
@@ -0,0 +1,300 @@
...
+    )
+
+    def setup(self) -> bool:
+        """Setup dnsperf benchmark."""
+        self.logger.info("Setting up dnsperf benchmark.")
+        self.config.parse_args()
+
+        if self.config.pod_id:
+            self.tmp_filepath += f"-{self.pod_id}"
+        if check_file(self.tmp_filepath):
```
learnitall, line 198: Is this tmp file used for something outside of the benchmark? I included this into the tutorial in the docs as an example for setup/cleanup tasks, but it isn't required.

snafu/benchmarks/dnsperf/dnsperf.py
```diff
@@ -0,0 +1,300 @@
...
+    def collect(self) -> Iterable[BenchmarkResult]:
+        """Run the dnsperf benchmark and collect results."""
+
+        self.logger.info("Starting dnsperf")
+        first_result: BenchmarkResult = self._one_trial()
+        yield first_result
+
+        max_qps = first_result.data["throughput"]
+        allowed_loads = (
+            np.linspace(
```
learnitall, line 276: Would it be possible to do this without numpy in order to save a dependency? The built-in ``range`` function includes a step parameter.

snafu/benchmarks/dnsperf/mocks/small-v.txt
```diff
@@ -0,0 +1,49 @@
+DNS Performance Testing Tool
```
learnitall, line 1: Is this file for testing?

learnitall (changes requested):
Other than a few minor details, looks great! Thanks for taking the time to learn the new benchmark format.

snafu/benchmarks/dnsperf/dnsperf.py
mfleader: The goal was to override the autogenerated repr method because it did not pretty print the `data` object..

snafu/benchmarks/dnsperf/dnsperf.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def collect(self) -> Iterable[BenchmarkResult]:
+        """Run the dnsperf benchmark and collect results."""
+
+        self.logger.info("Starting dnsperf")
+        first_result: BenchmarkResult = self._one_trial()
+        yield first_result
+
+        max_qps = first_result.data["throughput"]
+        allowed_loads = (int(max_qps * step / self.config.steps) for step in range(1, self.config.steps))
```
mfleader, line 219: I dislike how black formats this comprehension.

snafu/benchmarks/dnsperf/dnsperf.py
```diff
@@ -0,0 +1,234 @@
...
+            return
+        elif sample.successful.stdout is None:
+            self.logger.critical(
+                ("dnsperf ran successfully, but did not get output on stdout.\n" f"Got results: {sample}\n")
+            )
+            return
+
+        stdout: DnsperfStdout = self.parse_stdout(sample.successful.stdout)
+        cfg: DnsperfConfig = DnsperfConfig.new(stdout, self.config, load=load)
+        # self.logger.info(stdout)
```
mfleader, line 204: Should this info always logged to the terminal, or only when the user asks for verbose output?

snafu/benchmarks/dnsperf/dnsperf.py
mfleader: I thought we might want it for a multithreaded version that handles more than one pod. Other than that, I don't think we need it.

sjug:
This is impossible to review with @learnitall's PR mashed into this one.

snafu/benchmarks/dnsperf/Dockerfile
sjug: We can use centos stream when it's available. We can update the base image anytime.

snafu/benchmarks/dnsperf/Dockerfile
sjug: If you're not using them, then yes.

snafu/benchmarks/dnsperf/Dockerfile
mfleader: Holding Centos 8 for now, then.

snafu/benchmarks/dnsperf/dnsperf.py
learnitall: I'm voting for verbose output, that's my take

PR #305, opened 2021-07-20
State Signal Modules

Maxusmusti:
IMPORTANT NOTE:
ANY NEW FEEDBACK WOULD BE SEEN QUICKEST W/ AN ISSUE OR PR IN THE OFFICIAL REPO:
 - https://github.com/distributed-system-analysis/state-signals
 - This is where the most recent version of the code, documentation, releases, etc. lives now

------------------------------------------------------------------------------------------------------------------------------------------------------------------

EVENT/STATE SIGNAL MODULE    &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;         (NOW OPEN FOR REVIEWS!)

Adds two new, simple-to-use objects:
 - SignalExporter
 - SignalResponder

Also provides two dataclass specifications:
 - Signal
 - Response

Combining redis pubsub features with state signal + response protocols, these additions make state signal publishing, subscribing, receiving, and responding incredibly easy to integrate into any code. 

Anticipated use-case: benchmark runners will use the SignalExporter to publish benchmark state, and data collectors can subscribe and respond with the SignalResponder.

STILL TO-DO:
 - Question of how to name the responders (a.k.a. assign the tool_id). Should it be predictable so that expected subs can be added without the tool being up yet? Or unique so that a single host can have multiple responders for the same exporter? Or maybe there's a way to do both...


Maxusmusti:
See current design document [here](https://docs.google.com/document/d/1QpvW2qrjqGz4kL5b9iLjsUQ_pOR9KtcgEm89dCUlnZs/edit?usp=sharing)

signals/signal_exporter.py
```diff
@@ -0,0 +1,252 @@
...
+        self.redis = redis.Redis(host=redis_host, port=redis_port, db=0)
+        self.init_listener = None
+        self.legal_events = None
+
+    def _sig_builder(self, event, sample=-1, tag=None, metadata=None):
+        sig = Signal(publisher_id=self.pub_id, process_name=self.proc_name, event=event)
+        sig.runner_host = self.runner_host if self.runner_host else sig.runner_host
+        sig.tag = tag if tag else sig.tag
+        sig.metadata = metadata if metadata else sig.metadata
+        sig.sample_no = sample if sample >= 0 else sig.sample_no
```
dbutenhof, line 83: This "after the fact" defaulting is a bit awkward. At least for `sig_builder`, the effective defaults for `Signal` are really specified here rather than in the constructor.

Another way to override the defaults would be to build a kwarg dict here, omitting values where we don't have overrides, and then destructure that in the constructor call and leave `Signal` to manage its own defaults where we don't have explicit overrides.

dbutenhof (commented):
Just a bunch of random observations from reading through the code once ...

signals/signal_exporter.py
```diff
@@ -0,0 +1,252 @@
...
+        return sig
+
+    def _get_data_dict(self, response):
+        if not "data" in response:
+            print("No data in this response message")
+            return None
+        if isinstance(response["data"], int):
+            return None
```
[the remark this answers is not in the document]
Maxusmusti: There was a case I ran into with initial connection/sub messages that would have an int data to report success, but the more general form you suggested is a much better way to go about it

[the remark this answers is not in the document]
Maxusmusti: Oh wait actually it's not going to be a dict but a json string, so I can try/catch the json.loads() and return None on a Value error instead, and maybe also verify that its a string first to avoid relying solely on the try/catch (may hinder performance)

Maxusmusti:
Update: Will be changing 'tool_id' field to a new name (perhaps 'responder_id') for clarity. Will also be allowing user input for the responder_id, and will append an element of uniqueness as well. Thanks Joe for the helpful suggestions!

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+            return 3
+
+        sig = self._sig_builder(event=event, sample=sample, tag=tag, metadata=metadata)
+        sub_check, result_box = self._check_subs(event)
+
+        self.redis.publish(channel="event-signal-pubsub", message=sig.to_json_str())
+
+        counter = 0
+        while sub_check and sub_check.is_alive():
+            time.sleep(0.1)
```
learnitall, line 253: Is there a way that we can configure this to allow for setting a timeout value, rather than using the hard-coded 20 seconds?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+        Listen for responses from all registered subscribers. Return
+        listener, as well as value based on responders' RAS codes.
+        """
+        if not self.subs:
+            return None, [0]
+
+        to_check = set(self.subs)
+        subscriber = self.redis.pubsub(ignore_subscribe_messages=True)
+        result_box = [0]
```
learnitall, line 178: Are you using an array here to get around issues with scope?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+    def publish_signal(
+        self, event: str, sample: int = -1, tag: str = None, metadata: Dict = None
+    ) -> int:
+        """
+        Publish a legal event signal. Includes additional options to specify sample_no,
+        a tag, and any other additional metadata. Will then wait for responses from
+        subscribed responders (if any). Returns one of the below result codes based on 
+        signal publish/response success.
+
+        RESULT CODES:
```
learnitall, line 217: Can we add global constants to make this more readable? So up top define:

```python
ALL_SUBS_WELL = 0
SUB_FAILED = 1
MISSING_RESPONSE = 2
...
```

Might make the code in this function easier to read right off the bat

signals/signal_exporter.py
```diff
@@ -0,0 +1,252 @@
...
+            if not to_check:
+                listener.stop()
+
+        subscriber.subscribe(**{"event-signal-response": _sub_handler})
+        listener = subscriber.run_in_thread()
+        return listener, result_box
+
+    def _valid_str_list(self, names):
+        return (
+            bool(names)
```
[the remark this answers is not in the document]
learnitall: +1 on the type hint. Can use using ``typing.List`` rather than the built-in ``list``?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+    ) -> None:
+        """
+        Publishes an initialization message. Starts a listener that reads responses
+        to the initialization message and adds responders to the subscriber list.
+        Sets list of legal event names for future signals, and also allows for optional
+        input of expected hostnames (subscribers) as well as a tag.
+        """
+        if not self._valid_str_list(legal_events):
+            print("ERROR: 'legal_events' arg must be a list of string event names")
+            return
```
learnitall, line 272: Can we raise errors here rather than just return ``None``? There is nothing to differentiate failure in this function vs success.

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+            )
+            return 4
+
+        if event == "shutdown":
+            print("ERROR: Please use the 'shutdown()' method for shutdown signals")
+            return 5
+
+        if not skip_check and not event in self.legal_events:
+            print(f"Event {self.event} not one of legal events: {self.legal_events}")
+            return 3
```
learnitall, line 244: I feel like status codes 3, 4 and 5 should just be Python errors instead, as those are cases where the user isn't using the code correctly. Could we raise ``ValueError`` and/or ``TypeErrors`` here?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+        while sub_check and sub_check.is_alive():
+            time.sleep(0.1)
+            counter += 1
+            if counter >= 200:
+                print("Timeout after waiting 20 seconds for sub response")
+                return 2
+
+        return result_box[0]
+
+    def initialize(
```
learnitall, line 261: Can we add logs to print if initialize was successful or not?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+                return
+            for host in expected_hosts:
+                self.subs.append(host + "-resp")
+
+        self.legal_events = legal_events
+        sig = self._sig_builder(event="initialization", tag=tag)
+        self._fetch_responders()
+        self.redis.publish(channel="event-signal-pubsub", message=sig.to_json_str())
+
+    def shutdown(self, tag: str = None) -> None:
```
learnitall, line 286: Can we add logs to print if shutdown was successful or not?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def __init__(self, redis_host: str = "localhost", redis_port: int = 6379) -> None:
+        """
+        Sets exporter object fields and generates unique tool_id.
+        Allows for specification of redis host/port.
+        """
+        self.redis = redis.Redis(host=redis_host, port=redis_port, db=0)
+        self.subscriber = self.redis.pubsub(ignore_subscribe_messages=True)
+        self.subscriber.subscribe("event-signal-pubsub")
+        self.tool_id = platform.node() + "-resp"  # ADD UUID HERE?
```
learnitall, line 313: Can we allow for a custom string to be added in here to help with the debugging process? For instance, if I'm running two tools on the same node, I can't differentiate between them. Can we also add in the uuid before merge? If I'm running two tools on the same node, there will be a collision.

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+                "runner_host",
+                "sample_no",
+                "tag",
+                "metadata",
+            ]
+        )
+        if set(data.keys()) == check_set or check_set - set(data.keys()) == {
+            "metadata"
+        }:
+            if (not self.locked_id or self.locked_id == data["publisher_id"]) and (
```
learnitall, line 341: I totally understand what you are trying to do here, but can we revisit this? From my understanding here is the goal of the function:
* Validate that the signal is JSON
* Validate that the signal has all of the needed keys for constructing a signal.
* Filter out signals that aren't part of the ``locked_id`` or ``locked_tag``

What I'm confused about is:
* Why do we ignore data that are valid JSON, yet aren't valid signals, by returning ``None`` after the first if statement?
* Why don't we create a ``Signal`` instance from the given data and instead keep it as a ``dict``?
* The listen method is going to yield all of these ``None`` values when it gets signals to ignore, right? At least I think that's how it'll work.

So here is what I think would make this functionality really easy to extend and to read:

* Let ``_parse_signal`` be just a method that returns a ``Signal`` instance from the given data ``Dict``, raising ``ValueError`` if something is wrong.
* Create another method called ``_is_ignored`` or ``_is_target`` or something, which does the filtering by returning ``True/False`` on whether there is a match for ``locked_tag`` or ``locked_id`` when given a ``Signal`` instance
* Change ``listen`` to only yield when your ignore method says that the signal is a part of the locked tag and id.

This separates parsing and ignoring functionality into these bite-size pieces.

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+    def respond(self, publisher_id: str, event: str, ras: int = None) -> None:
+        """
+        Publish a legal response to a certain publisher_id's event signal.
+        Also allows for optional ras code to be added on (required for 
+        publisher acknowledgement, but not for initialization response).
+        """
+        response = Response(self.tool_id, publisher_id, event, ras)
+        self.redis.publish("event-signal-response", response.to_json_str())
+
+    def lock_id(self, publisher_id: str) -> None:
```
learnitall, line 366: Looks like these are setters for ``locked_id`` and ``lock_tag``, however users can still modify the ``locked_id`` and ``locked_tag`` directly. Python has the concept of properties for this purpose. Can we do one of these:

* Make ``locked_id`` and ``locked_tag`` properties, with a setter that does the validation
* Add underscores in front of ``locked_id`` and ``locked_tag`` to show that they shouldn't be modified by users directly

signals/signal_tester.py
```diff
@@ -0,0 +1,43 @@
+from signals import signal_exporter
```
learnitall, line 1: I really appreciate that you have this script here to show usage and for testing, but can we take it out of the branch? Maybe a make it a GitHub gist? I don't think it belongs in the source code. We can for sure add stuff like this when we get unit testing support.

learnitall (changes requested):
Really appreciate your work Mustafa, really awesome. Thank you for putting the time into this. I have a few implementation notes that I left that I think need to be addressed before merge, but this is awesome for sure. Really excited to get this in here. 

Here are some next steps and things that I think we should think about/discuss. I can help you with these items if need be, feel free to lmk offline:

* Would be a good idea to add some information in the docstrings on parameters and attributes for each of these classes, but not a big deal as your code is super readable
* Adding a page for this functionality into our online documentation (``docs/source/signals.md`` or something)
* Replacing print statements with logging functionality, as per your TODO list
* Maybe we could think about using something like [pydantic](https://pydantic-docs.helpmanual.io/)? @mfleader is using it for his PR #308 and it is friggin' awesome. It would help with the type validation that you have going on for your dataclasses.
* Getting your code to pass our pre-commit checks.

Thanks again Mustafa for this! Super pumped :)

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+
+        to_check = set(self.subs)
+        subscriber = self.redis.pubsub(ignore_subscribe_messages=True)
+        result_box = [0]
+
+        def _sub_handler(item: Dict) -> None:
+            data = self._get_data_dict(item)
+            if data and data["publisher_id"] == self.pub_id and data["event"] == event:
+                if "ras" in data:
+                    to_check.remove(data["tool_id"])
```
learnitall, line 184: This is going to raise a key error if the ``tool_id`` isn't in the ``to_check`` set. Do you think that we could add in a check like this:
```python
if data["tool_id"] not in to_check:
    print(f"WARNING: Got a response from tool '{data['tool_id']}' but it's not on the known subscribers list")
else:
    to_check.remove(data["tool_id"])
```

Let me know though if I'm not understanding this correctly and/or if this isn't needed.

learnitall (changes requested):
Forgot a comment, adding it in real quick here.

signals/signal_exporter.py
Maxusmusti: I think it actually handles that case by simply removing nothing, but your solution is still better as it significantly increases clarity.

Maxusmusti: The question is: if an unsubscribed tool responds with a bad response, do we still wanna give the warning that a tool returned a bad resp? Or would that not technically count as a "sub failing"? Really just a matter of whether or not to indent line 197 (see "non-sub handling" commit)

signals/signal_exporter.py
Maxusmusti: The issue is that stuff like metadata, sample_no, etc. can have no default in the exporter, since they are per-signal values. I could definitely try the kwarg dict though

signals/signal_exporter.py
Maxusmusti: Yeah, it's so that the listener can return a result code value that can be changed depending on responses

signals/signal_exporter.py
learnitall: I wish that sets would just remove nothing if the item wasn't in there, but it does raise a ``KeyError``:

```
>>> {"one", "two"}.remove("three")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'three'
>>> 
```

As a result I think it would be a good idea to add a check for this. I see what you're saying though about handling unsubscribed tools. Do you think then that we could do something like this:

```python
if data["tool_id"] in to_check: 
    to_check.remove(data["tool_id"])
```

And then add a print statement to log which tools we got a response from to help with debug?

signals/signal_exporter.py
```diff
@@ -0,0 +1,383 @@
...
+
+        self.redis.publish(channel="event-signal-pubsub", message=sig.to_json_str())
+
+        counter = 0
+        while sub_check and sub_check.is_alive():
+            time.sleep(0.1)
+            counter += 1
+            if counter >= 200:
+                print("Timeout after waiting 20 seconds for sub response")
+                return 2
```
learnitall: Just noticed this: do we need to stop the ``sub_check`` listener here? I'm concerned that it will keep working in the background even if we have encountered a timeout.

signals/signal_exporter.py
learnitall: Ah I see ok. If only Python had pointers right? My only trouble with this is that it is a bit hard to read for those not familiar with the intricacies of Python's atomic built-ins and the GIL that allow for a list like this to essentially work like a queue- I had to do quite a bit of Googling to understand how this worked. I know that this does work, and it's so cool that it does, but IMHO I would prefer if we used the built-in `queue` module. See Google's Python style guide: https://google.github.io/styleguide/pyguide.html#218-threading

signals/signal_exporter.py
learnitall: Ok I completely misunderstood your last comment and also just realized I was looking at outdated code :( my bad. What you pushed looks great. IMHO I think indenting the block at line 209 so we don't fail on unsubscribed responses would be a good idea. I'll trust your judgement on that though.

signals/signal_exporter.py
```diff
@@ -0,0 +1,421 @@
...
+        """
+        Listen for responses from all registered subscribers. Return
+        listener, as well as value based on responders' RAS codes.
+        """
+        if not self.subs:
+            return None, [0]
+
+        to_check = set(self.subs)
+        subscriber = self.redis.pubsub(ignore_subscribe_messages=True)
+        result_code_holder = [ResultCodes.ALL_SUBS_SUCCESS]
```
learnitall, line 197: Love what you did here, adding the ``_holder`` makes this easier to read through and understand what is going on. I did find a snippet from Google's Python style guide that warns on relying on the atomic operations of built-ins, however I think for this use case it should be all good: https://google.github.io/styleguide/pyguide.html#218-threading

signals/signal_exporter.py
Maxusmusti: We definitely do lol, nice catch

signals/signal_exporter.py
Maxusmusti: I agree, I think it doesn't make sense to fail on a non-sub. What I can do though is in the warning for the non-sub also display its ras, just so the info is available

signals/signal_tester.py
Maxusmusti: Yeah, I'll manage the script organization and documentation and testing and stuff once I've figured out where this work will actually ending up living. It may end up staying in this repo, or it may end up in its own repo (I'll prolly ask abt it at the tools meeting this week)

signals/signal_exporter.py
```diff
@@ -0,0 +1,252 @@
...
+        def _init_handler(item):
+            data = self._get_data_dict(item)
+            if (
+                data
+                and data["event"] == "initialization"
+                and data["publisher_id"] == self.pub_id
+            ):
+                self.subs.append(data["tool_id"])
+
+        subscriber.subscribe(**{"event-signal-response": _init_handler})
```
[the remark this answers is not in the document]
Maxusmusti: Could you explain what changes when switching from hyphens to underscore? This was my initial reason for using the current dict: 
image

signals/signal_exporter.py
Maxusmusti: When I created the signal and response dataclasses, I had initially thought of them as internal objects, though you raise a fair point. Do you think it would be better for the user to receive a Signal instead of a dict with the fields? If so I can make that change, and in the meantime I'll also separate out the componentsof this process

Maxusmusti: Fixed everything else but returning a signal (an easy change though). Curious to see if you think returning a Signal would be more convenient or confusing for a user.

mfleader: Is the user intended to be only other Python programs, or any application that can consume a JSON?

learnitall: Awesome, thank you for taking the time to do that change. I think @mfleader hit it right on the money with this question. I think that it would be safe to assume that any user using the Python library would be expected to use the Python dataclasses rather than a JSON dict, whereas we can leave the JSON dict usage to other non-python users of the protocol. Additionally, I like the more OO approach of using the signal class rather than the dict. Makes the returned data easy to document and read from a user perspective IMHO.

signals/signal_exporter.py
```diff
@@ -0,0 +1,483 @@
...
+
+
+def _create_logger(
+    class_name: str, process_name: str, log_level: str
+) -> logging.Logger:
+    """
+    Creates and returns logging.Logger object for detailed logging.
+    Used by SignalExporter and SignalResponder.
+    """
+    logger = logging.getLogger(class_name).getChild(process_name)
```
learnitall, line 36: Do you think we could modify this to use a base logger with a specific name for the signals? As well as use a global log-level for the loggers that makes the messages easy to control? Something like:

```python

def setup_logging():
    base_logger = logging.getLogger("state-signals")
    stream_handler = logging.StreamHandler()
    try:
        base_logger.setLevel(log_level)
        stream_handler.setLevel(log_level)
     except ValueError:
          ....
        
    formatter = ....
    stream_handler.setFormatter(formatter)
    base_logger.addHandler(stream_handler)
    
    return base_logger

_base_logger = _get_base_logger()

def _create_logger(class_name: str, process_name: str, log_level: str) -> logging.Logger:
    """..."""

    return _base_logger.getChild(f"{class_name}.{process_name}")
```

Or even just:

```python
def _create_logger(...):
    logger = logging.getLogger("state-signals").getChild(class_name).getChild(process_name)
    ...
```

This would establish the standard practice logging hierarchy as well as make these log messages easy to control from other apps. For instance, I could mute signal logs in my app and redirect to a file with:

```python
import logging
import state_signals

logging.getLogger("state-signals").setLevel(logging.ERROR)
logging.getLogger("state-signals").addHandler(logging.FileHandler("/tmp/state-signals.log"))
```

Or whatever other sort of logging configuration my app needs.

learnitall (commented):
Left some quick comments on the new logging stuff that you added.

signals/signal_exporter.py
```diff
@@ -0,0 +1,483 @@
...
+    legal signals and handling responders.
+    """
+
+    def __init__(
+        self,
+        process_name: str,
+        redis_host: str = "localhost",
+        redis_port: int = 6379,
+        runner_host: str = platform.node(),
+        log_level: str = "INFO",
```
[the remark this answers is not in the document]
Maxusmusti: I felt like leaving it all as strings (including the default) would be easier to read/understand for the user. That way they don't need any prior knowledge of the logging module.

learnitall:
Migrated to [distributed-system-analysis/state-signals](https://github.com/distributed-system-analysis/state-signals)

PR #302, opened 2021-07-19
Bump es version

learnitall:
### Description

This PR bumps elasticsearch client version within our ``setup.cfg`` to use ``elasticsearch>=7.0.0,<8.0.0``, (as per documentation spec at https://elasticsearch-py.readthedocs.io/en/v7.14.0a1/), in order to enable Python 3.9 support.  Joe and I discussed that we would need to test this change against our ES instances to make sure bumping the client version doesn't break our data ingest.

### Fixes

Currently we have elasticsearch client pinned as ``elasticsearch>=6.0.0,<=7.0.2``, however these versions are not compatible with Python 3.9. The following ``ImportError`` is observed when trying to invoke snafu on Python 3.9:

```
snafu/run_snafu.py:20: in <module>
    import elasticsearch
.tox/py39/lib/python3.9/site-packages/elasticsearch/__init__.py:24: in <module>
    from .client import Elasticsearch
.tox/py39/lib/python3.9/site-packages/elasticsearch/client/__init__.py:4: in <module>
    from ..transport import Transport
.tox/py39/lib/python3.9/site-packages/elasticsearch/transport.py:4: in <module>
    from .connection import Urllib3HttpConnection
.tox/py39/lib/python3.9/site-packages/elasticsearch/connection/__init__.py:2: in <module>
    from .http_requests import RequestsHttpConnection
.tox/py39/lib/python3.9/site-packages/elasticsearch/connection/http_requests.py:3: in <module>
    from base64 import decodestring
E   ImportError: cannot import name 'decodestring' from 'base64' (/opt/hostedtoolcache/Python/3.9.6/x64/lib/python3.9/base64.py)
```

Additionally, I removed the ``elasticsearch1`` module import, as we only have ES >= 7 in our environment, which is supported within the given ``elasticsearch`` module itself.

PR #306, opened 2021-07-26
Add Unit and Documentation Build Tests

learnitall:
### Description

Adds functionality for unit testing by leveraging pytest and tox. I believe that our goal for unit testing with benchmark-wrapper is to test the core functionality that is shared among our benchmark wrappers, as testing the wrappers themselves should be left to functional testing.

Here is how it works:

* Create a new test module under ``tests/unit/test_<module>.py``, where ``<module>`` is the name of a core snafu module.
* Write unit tests within the new test module using pytest
* Optionally, [write docstring tests](https://docs.pytest.org/en/6.2.x/doctest.html) for functions/classes within then core snafu module
* Invoke with tox
* Review the printed coverage report.

I've configured tox with eight different environments, allowing us to test across four versions of Python (3.6-3.9) and against unit and documentation tests: ``py{36,37,38,39}-{unit,docs}``. For example, to run unit tests against Python 3.6, run: ``tox -e py36-unit``. To run documentation build tests against Python 3.8, run ``tox -e py38-docs``. Running tox without any arguments will tell tox to attempt to run all eight environments, ignoring any unavailable Python versions.

[Coverage](https://coverage.readthedocs.io/en/coverage-5.5/) reports are saved as ``.coverage.<tox environment>``. Configuration file is under ``.coveragerc`` and adds ignores for all benchmark-related code.

Additionally, I've added GH Actions workflows with will automatically run these tests on pull-requests, pushes or on a manual trigger. Coverage reports from the unit test workflow will automatically be uploaded to [codecov](https://docs.codecov.com/docs/), so we can share and review coverage reports in detail.


learnitall:
I haven't included Python 3.9 in the GHA workflows due to this bug: https://github.com/cloud-bulldozer/benchmark-wrapper/pull/302

whitleykeith (approved):
lgtm

docs/source/contributing/unit_testing.md
```diff
@@ -0,0 +1,38 @@
+# Testing
+
+benchmark-wrapper uses [tox](https://pypi.org/project/tox/) and [pytest](https://docs.pytest.org/en/6.2.x/) for unit testing as well as documentation build testing.
+
+As a quick reminder, unit testing is defined as follows by [STF](https://softwaretestingfundamentals.com/unit-testing/):
+
+> UNIT TESTING, also known as COMPONENT TESTING,  is a level of software testing where individual units / components of a software are tested. The purpose is to validate that each unit of the software performs as designed.
```
dry923, line 7: extremely minor nit but there's 2 spaces after the second ,
i.e.
COMPONENT TESTING,  is....

learnitall: Woah nice catch. That's funny as that's a copy-paste from https://softwaretestingfundamentals.com/unit-testing/. I removed the space though and will commit and push

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,30 @@
+name: Unit Tests
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+  pull_request:
+  workflow_dispatch:
+
+jobs:
```
rsevilla87, line 9: Maybe we can use a matrix here to speed-up the workflow?

learnitall: I've figured out how to do this, about to push, but one thing I'm concerned about with this is that the UI will show all of the jobs and it can get kinda messy with 8+ jobs. Especially since the functional tests will have a job for each benchmark, this may get crazy! If it's something we are ok with though then that's great!

mohit-sheth (approved):
LGTM

tests/unit/test_config.py
```diff
@@ -0,0 +1,133 @@
...
+        assert config.param4 == 4321
+
+    @staticmethod
+    def get_config_instance() -> snafu.config.Config:
+        """Return a new Config instance with tool_name set to 'TEST'."""
+        try:
+            configargparse.init_argument_parser()
+        except ValueError:
+            del configargparse._parsers["default"]  # pylint: disable=W0212
+            configargparse.init_argument_parser()
```
rsevilla87, line 80: Why do we need these lines?

learnitall: [configargparse](https://pypi.org/project/ConfigArgParse/) uses a global argument parser which we are taking advantage of, but in order to have a clean parser for each test we need to make sure that we refresh it, and that's what these lines do. Here are the relevant lines in configargparse: https://github.com/bw2/ConfigArgParse/blob/master/configargparse.py#L35

.github/workflows/unit.yaml
```diff
@@ -0,0 +1,50 @@
...
+          python-version: 3.${{ matrix.pyver }}
+      - name: Install tox
+        run: pip install tox
+      - name: Run Tox
+        run: tox -e py3${{ matrix.pyver }}-unit
+      - name: Upload coverage report artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: coverage-py3${{ matrix.pyver }}
+          path: .coverage.py3${{ matrix.pyver }}-unit
```
rsevilla87, line 29: I think we can skip uploading these artifacts. Coverage will be the same regardless of the python version. We should trust the coverage_report  job for that.

learnitall: Ah ok good point. I think the example that I looked at might have had both Python 2 and 3 support where this kind of combined coverage would need to be leveraged. Since we are assuming that are code will work across each of the four Python 3 versions that we support, as the unit tests stay the same across the versions, I'll simplify this workflow.

rsevilla87 (approved):
LGTM!

PR #307, opened 2021-07-27
Change linpack URL to latest Intel bits, change LINPACK_DAT so no need to be root

mnmehta:
### Description

Move to 2021 version of Intel linpack

### Fixes

- Updated linpack URL, old URL is no longer reachable
- Also changed LINPACK_DAT to be in /tmp instead of /opt, so it will run on openshift without needing a privileged pod
- Interface remains unchanged

### DIFF
--- a/snafu/linpack_wrapper/Dockerfile
+++ b/snafu/linpack_wrapper/Dockerfile
@@ -6,5 +6,5 @@ RUN mkdir -p /opt/snafu/
 COPY . /opt/snafu/
 RUN pip3 install -e /opt/snafu/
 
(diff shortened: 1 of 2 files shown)


learnitall:
This looks great, thank you!

/rerun all

learnitall:
Looks like we don't have CI tests for linpack, but I can verify that this builds successfully and I can run our ``linpack.sh`` script. From my end LGTM. @jtaleric or @mohit-sheth do you mind taking a look?

jtaleric (approved):
LGTM - The issue with Linpack is we will have to maintain the tarball location with each container build. I wonder if there is a better way?

PR #309, opened 2021-07-27
Fix sysbench's Dockerfile

ewchong:
### Description

The Dockerfile for sysbench was missing benchmark-wrapper and wouldn't build for AARCH64.

### Fixes

- Install benchmark-wrapper in sysbench's Dockerfile
- Update pip3 inside the container so it will build on a Raspberry PI4

### DIFF
--- a/snafu/sysbench/Dockerfile
+++ b/snafu/sysbench/Dockerfile
@@ -7,5 +7,6 @@ RUN dnf install -y --nodocs sysbench && dnf clean all
 RUN dnf install -y --nodocs python3-pip procps-ng iproute net-tools ethtool nmap iputils && dnf clean all
 RUN ln -s /usr/bin/python3 /usr/bin/python
 COPY . /opt/snafu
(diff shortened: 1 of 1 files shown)


learnitall (commented):
Thanks for taking the time to do this! Can you move the snafu install steps after the sysbench install steps? Will help speed up builds in the future as we start to take advantage of layer caching.

ewchong:
> Thanks for taking the time to do this! Can you move the snafu install steps after the sysbench install steps? Will help speed up builds in the future as we start to take advantage of layer caching.

Done. I was thinking that eventually benchmark-wrapper could have its own Dockerfile and that could be the base image. Would get rid of a lot of repetitions in the other Dockerfiles and speed up the builds.

learnitall:
Looks great, thank you! On the base images, that's definitely something that we have in the works. We recently did a review of benchmark-wrapper and am right now going through a "modernization" period of changes.

PR #310, opened 2021-07-27
Fix sysbench path

ewchong:
### Description

On Ubuntu, sysbench only exists on `/usr/bin`, on CentOS 8 it is in both `/bin/` and `/usr/bin`

### Fixes

- Look for sysbench in `/usr/bin/sysbench`


learnitall:
Thanks for the catch. LGTM. @mohit-sheth or @jtaleric can one of y'all take a look?

PR #312, opened 2021-08-04
Remove unused podman push

rsevilla87:
### Description
We don't need to `podman push`anymore


dry923 (approved):
LGTM!

PR #297, opened 2021-07-14
Proper parsing for uperf connect type workload

josecastillolema:
### Description
uperf connect type workloads do not use the `Txn2` output, only `Txn1`

PR #314, opened 2021-08-05
Fix Uperf Parsing of Connect Type Workload (via #297)

learnitall:
### Description

Applies PR #297 onto the new uperf code. From @josecastillolema in #297:

"uperf connect type workloads do not use the Txn2 output, only Txn1"


jtaleric (approved):
lgtm

PR #316, opened 2021-08-05
fix sysbench docker for aarch64

ewchong:
### Description

2nd round of fixes to get Docker image building on a Raspberry PI 4.

### Fixes

- Was missing `pip` when I did #309 for the upgrade
- Failed to install benchmark-wrapper since it requires at least Python 3.7


learnitall:
Ah nice find! Looks like scipy recently deprecated support for Python 3.6 ([source](https://github.com/scipy/scipy/releases/tag/v1.7.0)) and benchmark-wrapper's setup.cfg is out of date. This seems to be a bigger problem than just the sysbench wrapper. LGTM on the changes though for sure.

PR #319, opened 2021-08-09
Add Automated Image Build System

learnitall:
### Description

Goal of this PR is to add an automated container build mechanism that will build each benchmark image across different architectures and publish to quay, all through GHA.


learnitall:
Added in build workflow. Will need to set the following secrets:

* ``secrets.REGISTRY_USER``: User or org where we want images published (i.e. cloud-bulldozer)
* ``secrets.REGISTRY_AUTH_USER``: Username for authenticating to quay
* ``secrets.REGISTRY_PASSWORD``: Password for authenticating to quay

ci/build_matrix.py
```diff
@@ -0,0 +1,303 @@
...
+        return False
+
+    def build(self):
+        """Build the GHA jobs matrix."""
+
+        bones_changed = self.bones_changed()
+        for dockerfile in self.dockerfile_set:
+            changed = bones_changed or self.benchmark_changed(dockerfile)
+            entry = MatrixEntry.new(dockerfile=dockerfile, archs=self.archs, changed=changed,)
+            self.add_entry(entry)
```
rsevilla87, line 286: Rather than adding an entry for changed/not-changed benchmarks (AKA dockerfiles), why don't we generate a list with only changed benchmarks?, that way we could skip the parameter changed and generate a more user friendly output :).

```python
            if bones_changed or self.benchmark_changed(dockerfile):
                entry = MatrixEntry.new(dockerfile=dockerfile, archs=self.archs)
                self.add_entry(entry)
```

learnitall: Implemented! I added a new option to the CLI `--changed-only` which will only output changed Dockerfiles. I want to keep in the ability to output Dockerfiles that haven't changed as well, in order to support a manual re-build of all images for testing/debug purposes.

learnitall:
Can successfully run builds against all images or against just changes. Here are examples:

* Run build on all images regardless of changes: https://github.com/learnitall/benchmark-wrapper/runs/3374201758?check_suite_focus=true
* Run build on only changed images: https://github.com/learnitall/benchmark-wrapper/runs/3374590871?check_suite_focus=true
  * For this build, I created a new branch off of this PR called `bug-fix-black` and manually triggered a build, using this PR as the upstream branch to compare changes to. Since only `dns_perf` and `smallfile` wrappers were changed, only those images were built.

rsevilla87 (approved):
Lgtm

learnitall:
Here is a diagram I made on the workflow:

Snafu GHA Docker Build Workflow

whitleykeith (approved):

learnitall:
Failed doc tests on link check. I added in links to files within our repo on GitHub that are contained within this PR. Since this isn't merged yet, the links don't exist yet. Ignoring these errors, as they are expected.

PR #321, opened 2021-08-10
upgrade comparison patch

mukrishn:
### Description
[This](https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/snafu/upgrade_openshift_wrapper/trigger_upgrade.py#L218) validation is not right, `desired_version` variable from [here](https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/snafu/upgrade_openshift_wrapper/trigger_upgrade.py#L191) has got the old version prior to upgrade and validating with that cause the script to fail even after a successful upgrade. Instead it should compare with version post upgrade command. 

```
2021-08-09T19:53:48Z - INFO     - MainProcess - trigger_upgrade: Cluster upgrade complete
...
2021-08-09T19:53:48Z - ERROR    - MainProcess - trigger_upgrade: Cluster did not upgrade to desired version
2021-08-09T19:53:48Z - ERROR    - MainProcess - trigger_upgrade: Cluster version is 4.7.16 and desired version is 4.7.13
```
Updated the var scope fixes it, 
```
2021-08-10T18:04:19Z - INFO     - MainProcess - trigger_upgrade: Finished upgrading the cluster to version 4.7.13
```

### DIFF
--- a/snafu/upgrade_openshift_wrapper/trigger_upgrade.py
+++ b/snafu/upgrade_openshift_wrapper/trigger_upgrade.py
@@ -116,7 +116,15 @@ class Trigger_upgrade:
             the_time = datetime.datetime.strptime(
                 clusterversion.get().attributes.items[0].status.history[0].completionTime, time_format
             )
-            return init_version, desired_version, platform, the_time, the_time, (the_time - the_time)
+            return (
+                init_version,
+                desired_version,
+                desired_version,
+                platform,
+                the_time,
+                the_time,
+                (the_time - the_time),
+            )
 
         c_state = "incomplete"
         c_version = "0.0.0"
@@ -161,7 +169,7 @@ class Trigger_upgrade:
         total_time = end_time - start_time
         logger.info("Total upgrade time: %s" % total_time)
 
-        return init_version, new_version, platform, start_time, end_time, total_time
+        return init_version, desired_version, new_version, platform, start_time, end_time, total_time
 
     def get_timings(self):
         logger.info("Getting timing stats from cluster Operators")
@@ -199,6 +207,7 @@ class Trigger_upgrade:
         self.timestamp = time.strftime("%Y-%m-%dT%H:%M:%S")
         (
             self.init_version,
+            self.desired_version,
             self.end_version,
             self.platform,
             self.start_time,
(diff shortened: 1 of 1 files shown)


snafu/upgrade_openshift_wrapper/trigger_upgrade.py
```diff
@@ -161,7 +161,7 @@ def _run_upgrade(self):
         total_time = end_time - start_time
         logger.info("Total upgrade time: %s" % total_time)
 
-        return init_version, new_version, platform, start_time, end_time, total_time
+        return init_version, desired_version, new_version, platform, start_time, end_time, total_time
```
learnitall, line 164: This return statement was updated, but the return statement on line 119 was not. Can we update that one as well? Looks like it would need to be:

```python
return init_version, desired_version, desired_version, platform, the_time, the_time, (the_time - the_time)
```

snafu/upgrade_openshift_wrapper/trigger_upgrade.py
```diff
@@ -199,6 +199,7 @@ def emit_actions(self):
         self.timestamp = time.strftime("%Y-%m-%dT%H:%M:%S")
         (
             self.init_version,
+            self.desired_version,
```
learnitall, line 202: Does the log statement on line 195 output the correct desired version? Based on your changes I don't think it will and that now it needs to be updated?

learnitall (changes requested):
Left a few comments on the changes. Can you take a look and additionally do a run of pre-commit? Thanks!

snafu/upgrade_openshift_wrapper/trigger_upgrade.py
mukrishn: yes my bad, that was missed, corrected it now. Thanks

snafu/upgrade_openshift_wrapper/trigger_upgrade.py
mukrishn: Log line 195 prints a different variable, thats set in 194. It is an output from the cluster itself.

learnitall: Ah ok. I was confused because I thought the output from the cluster itself would be inaccurate since it seemed like the upgrade hadn't been triggered yet. My bad!

PR #311, opened 2021-08-04
Downgrade elasticsearch-py version due to incompatibility issues

rsevilla87:
Signed-off-by: Raul Sevilla <rsevilla@redhat.com>

### Description

It seems like elasticsearch-py is not supporting anymore ES "forks" like the AWS one starting from the version 7.14
Some context:

https://github.com/elastic/elasticsearch-py/blob/master/docs/guide/release-notes.asciidoc#7140-2021-08-02
https://github.com/elastic/elasticsearch-py/issues/1639#issuecomment-883319286

### Fixes
Fixes ES indexing in all workloads
```python
>>> es=elasticsearch.Elasticsearch("https://foo:bar@search-perfscale-pro-asdfasdfasdf.us-west-2.es.amazonaws.com")
>>> es.ping()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/elasticsearch/client/utils.py", line 168, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/elasticsearch/client/__init__.py", line 282, in ping
    "HEAD", "/", params=params, headers=headers
  File "/usr/local/lib/python3.6/site-packages/elasticsearch/transport.py", line 413, in perform_request
    _ProductChecker.raise_error(self._verified_elasticsearch)
  File "/usr/local/lib/python3.6/site-packages/elasticsearch/transport.py", line 630, in raise_error
    raise UnsupportedProductError(message)
elasticsearch.exceptions.UnsupportedProductError: The client noticed that the server is not a supported distribution of Elasticsearch
```

PR #323, opened 2021-08-12
Add Requirement Files to Pin Dependency Versions

learnitall:
### Description

Adds requirement files to pin versions for our dependencies, which will help to catch and resolve issues such as the one found in https://github.com/cloud-bulldozer/benchmark-wrapper/pull/316, or https://github.com/cloud-bulldozer/benchmark-wrapper/pull/311

The idea is to use [pip-tools](https://github.com/jazzband/pip-tools) to pre-compile a requirements file for each Python version, then use said file for install. This will create reproducible builds across our CI systems and for our users, preventing dependency resolution issues from unexpectedly impacting workloads.

This PR also adds:

* Tox environments for running pip-compile for each Python version we support.
* Integration of the requirement files within our documentation and unit testing.
* GHA workflow to perform the dependency resolution on changes to `setup.cfg` and once a week on Mondays. If any changes to dependency versions are detected, then a PR is automatically submitted with the changes which we can review for merge. Here is an example of what this PR looks like: https://github.com/learnitall/benchmark-wrapper/pull/23

### Fixes

Prevents upgrades in our dependencies from breaking our installs, as well as helps us keep support for Python 3.6 as dependent packages deprecate support. For instance, one can see that the versions for scipy and numpy differ between Python 3.6 and 3.7.


learnitall:
Closing in favor of implementing this solution with python-poetry. Will be much easier to maintain and much less complex.

learnitall:
Alright, this problem is really hard for me. To summarize, the problem that I'm trying to solve is to add some way for us to prevent unexpected dependency resolution problems from breaking our Docker image builds. This article provides great context on this problem: https://pip.pypa.io/en/latest/topics/dependency-resolution/#possible-ways-to-reduce-backtracking. Here are the needs:

We need a solution that:

* Will keep our Docker images small (low overhead)
* Creates minimal disruption to other teams and their workflows (small changes)
* Prevents lock-in (adaptable)
* Will work in the long term (maintainable)

Here is why poetry doesn't work:

* Have to use poetry for install with the lock file, which increases installation size and time. We need to install poetry dependencies then we can install our package. We could have poetry output a requirements file that pip can then use, but the whole point of using poetry is for leveraging its lock file and we'd have to add some GHA workflow for this anyways.
* The lock file updates whenever someone uses `poetry add`, the standard CLI command to add a new dependency, which means that it's easy for this lock file to be accidentally updated by a user.
* Based on how the lock mechanism works, we'd need to either deprecate python 3.6 support, requiring installing a new version of Python in CentOS based environments which ship 3.6 by default, or we need to figure out which versions our dependencies need to be pinned to in order to support python 3.6. We'd then have to do the same dependency pinning process for Python 3.7 in a year when it gets phased out. This just moves our work from troubleshooting builds to pinning dependencies

A good discussion on this was found by @mfleader here: https://hynek.me/articles/python-app-deps-2018/

These are the reasons why I think this branch will actually work and is the way to go. There is a lot of repeated lines of text, but it's the easiest way to have a proper long-term solution that will meet our needs and keep our builds stable. Reopening.

mfleader (changes requested):
This is legit. Good job, @learnitall. 

Does this affect the README's instructions on how to install benchmark wrapper?

learnitall:
Yes it does, good catch. I'll update docs and such.

learnitall:
And thank you! :)

mohit-sheth (approved):
lgtm

whitleykeith (approved):

learnitall:
Will update docs to reflect changes then merge!

PR #324, opened 2021-08-16
Stressng: add Elasticsearch fields for visualization

ebattat:
### Description
Stressng: Adding ElasticSearch fields for visualization ( This PR is depend on the same benchmark-operator PR https://github.com/cloud-bulldozer/benchmark-operator/pull/641 )
### Fixes
Adding the following elasticsearch fields:
es_ocp_version
es_cnv_version
es_vm_os_version
es_rhcos_version
es_kata_version
es_kind
es_data

### DIFF
--- a/snafu/stressng_wrapper/stressng_wrapper.py
+++ b/snafu/stressng_wrapper/stressng_wrapper.py
@@ -30,6 +30,7 @@ class stressng_wrapper:
         self.args.es_rhcos_version = ""
         self.args.es_kata_version = ""
         self.args.es_kind = ""
+        self.args.es_build_version = ""
         self.args.es_data = ""
 
         if "runtype" in os.environ:
@@ -55,6 +56,8 @@ class stressng_wrapper:
             self.args.es_kata_version = os.environ["es_kata_version"]
         if "es_kind" in os.environ:
             self.args.es_kind = os.environ["es_kind"]
+        if "es_build_version" in os.environ:
+            self.args.es_build_version = os.environ["es_build_version"]
         if "es_data" in os.environ:
             self.args.es_data = os.environ["es_data"]
 
--- a/snafu/stressng_wrapper/stressng_wrapper.py
+++ b/snafu/stressng_wrapper/stressng_wrapper.py
@@ -23,6 +23,14 @@ class stressng_wrapper:
         self.args.vm_stressors = ""
         self.args.vm_bytes = ""
         self.args.mem_stressors = ""
+        # es custom fields
+        self.args.es_ocp_version = ""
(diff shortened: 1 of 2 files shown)


jtaleric (changes requested):
Please review what we use `metadata-collector`[1].

[1] https://github.com/cloud-bulldozer/metadata-collector

rsevilla87:
@ebattat The fields you are adding is what we call metadata. We shouldn't include all the metadata in the benchmark documents as the indexed data will eventually become unmanageable. As @jtaleric mentioned, you can use metadata-collector; which actually leverages stockpile to collect metadata. https://github.com/cloud-bulldozer/stockpile/. Find more info about how benchmark-operator orchestrates metadata collection at https://github.com/cloud-bulldozer/benchmark-operator/blob/master/docs/metadata.md#how-it-works

I'd suggest you to create a new generic ansible role in stockpile for CNV/Kata metadata. IIRC metadata collection is not enabled in workloads based on VMs, so I think we should start working from that point.

rsevilla87:
Using kubevirt's hook sidecars can be used to trigger metadata collection from a VMI.
https://github.com/kubevirt/kubevirt/blob/main/examples/vmi-with-sidecar-hook.yaml#L6-L7

ebattat:
@rsevilla87 , @jtaleric 
https://github.com/cloud-bulldozer/metadata-collector/stockpile is really good tool but it use a separate **metadata index**.
We must get the metadata in the workload index for creating informative/appropriate visualizations per workload run.
Do you have a way/tool to upload the metadata into the **workload index** ?

bengland2:
I agree with Raul.   why is it necessary to put all this in the workload index?   I prefer to separate out different types of documents into different indexes.   This makes it easier to find and look at the data with either scripts, grafana or custom python scripting.   For example, backpack (stockpile) data can be found in the same index regardless of which benchmark generated it!   The test uuid is the cross-reference that connects the data in different indexes.

jtaleric:
> @rsevilla87 , @jtaleric
> https://github.com/cloud-bulldozer/metadata-collector is really good tools but it use a separate **metadata index**.
> We must get the metadata in the workload index for creating informative/appropriate visualizations per workload run. (i.e. get the workload kind vm/pod in order to share the results of pod and vm on the same dashboard)
> Do you have a different way/tool to upload custom metadata into the **workload index** ?

So, if you need to have the data all together, I propose you have a post task to pull specific meta you need, and index into a combined ES index. 

The issue with adding this arbitrary data is that, you need these fields today, I am sure you will need more later. Better just let the result index be results, and metadata be separate. IMHO. Happy to hear your thoughts on this.

PR #328, opened 2021-08-23
common code for redis pod synchronization

bengland2:
Do not merge yet.
storage benchmarks may have to synchronize pods 2 times per sample, 1 time after cache drop and 1 time after benchmark run
this PR creates common code to simplify pod synchronization to avoid duplication across benchmarks
we try it out with smallfile benchmark to illustrate its use but fio, fs-drift and someday cosbench could use it too.


bengland2:
tested using 30 smallfile pods with 1 threads each, no problems encountered, was doing both kernel and Ceph OSD cache dropping at same time, you can see other pods waiting for client-1 to finish cache dropping in log.   Will merge once tests complete.

PR #330, opened 2021-08-24
Remove LiveProcess Wrapper

learnitall:
### Description

This PR removes the LiveProcess wrapper from snafu. LiveProcess was added in #283 in an effort to allow for parallel stdout parsing and benchmark execution. As @mfleader pointed out to me, and discovered during his performance testing with dnsperf, this is an async approach that is implemented in a weird, non-pythonic way with multi-threading. We've decided to remove LiveProcess for now and, when the time is right, implement async execution within snafu using the standard libraries. This will reduce our overall code maintenance requirements.

### Fixes

* High resource usage of benchmark-wrapper with very verbose benchmarks
* Deadlock scenario when OS process pipe becomes full (https://docs.python.org/3/library/subprocess.html#popen-objects)


snafu/process.py
```diff
@@ -30,152 +28,7 @@ class ProcessSample:
...
-            self.attempt.rc = self.process.returncode
-            self.attempt.time_seconds = (self.end_time - self.start_time).total_seconds()
-
-    def __exit__(self, *args, **kwargs):
-        """Call cleanup method on exit of context."""
-        self.cleanup()
+    successful: Optional[ProcessRun] = None
 
 
 def get_process_sample(
```
mfleader, line 34: Remove mention of LiveProcess in doc string.
```
    """
    Run the given command as a subprocess, retrying if the subprocess fails, returning a
    :py:class:`~snafu.process.ProcessSample` detailing the results.

    This function expects a logger because it is expected that it will be used by benchmarks, which should
    be logging their progress anyways.

    Parameters
    ----------
    cmd : str or list of str
        Command to run. Can be string or list of strings if using :py:mod:`shlex`
    logger : logging.Logger
        Logger to use in order to log progress.
    retries : int, optional
        Number of retries to perform. Defaults to zero, which means that the function will run the process
        once, not retrying on failure.
    expected_rc : int, optional
        Expected return code of the process. Will be used to determine if the process ran successfully or not.
    timeout : int, optional
        Time in seconds to wait for process to complete before killing it.
    kwargs
        Extra kwargs will be passed to :py:class:`~snafu.process.ProcessSample`

    Returns
    -------
    ProcessSample
    """
```

snafu/process.py
```diff
@@ -223,13 +76,41 @@ def get_process_sample(
     tries: int = 0
     tries_plural: str = ""
```
mfleader, line 78: What is the purpose of `capture_output`? Is this for backwards compatibility?

mfleader (changes requested):
should process.get_process_sample() be a private function? 

It seems like the intended entrypoint is process.sample_process(), and if I only want to do one execution of a command, then I should use process.sample_process(num_samples=1).

learnitall:
Good point @mfleader , there seems to be too much overlap between `get_process_sample` and `sample_process`. Let me adjust so it makes more sense. I want `get_process_sample` to be a function that can just run a process a single time, I don't want to make it private.

snafu/process.py
learnitall: Yeah exactly. The kwarg isn't available in 3.6, so it needs to be removed before being given to `subprocess.run`.

snafu/process.py
learnitall: Thank you for the catch!

learnitall:
I just took a look at the code and I think it's a good idea to keep `get_process_sample` as is. I was mistaken: the goal of `get_process_sample` is to run a process _with retries_. `sample_process` will sample a process multiple times, retrying each sample as needed. They are pretty distinct. For instance, if I want to run a process to do a setup task, I'd want to use `get_process_sample` in order to do so as it will be retried as needed.

mfleader:
So it is for users that want to run a process multiple times, and need to re-run their setup every time they execute?

mfleader (approved):

PR #331, opened 2021-08-25
Fix Failing pre-commit Hooks

learnitall:
### Description

The pre-commit CI is failing for our repository, as we had some drift between our code and our hooks. I ran our hooks across our code base using `pre-commit run --all` and added the changes in this PR. Some changes were manually made in order to fix flake8 linting issues, however this only involved renaming variables used within list comprehensions. The code should still function exactly the same.

### Fixes

Failing pre-commit CI runs.

### DIFF
--- a/snafu/cyclictest_wrapper/cyclictest_wrapper.py
+++ b/snafu/cyclictest_wrapper/cyclictest_wrapper.py
@@ -11,8 +11,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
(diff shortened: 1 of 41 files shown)


learnitall:
TIL always google an issue before bothering moderators. Using `--install-types` is unsupported in pre-commit CI as environments are meant to be immutable. Will use [additional_dependencies](https://pre-commit.com/#config-additional_dependencies) arg to install the needed type packages

PR #336, opened 2021-08-31
IOPS floating-point truncation to int results in significant fio errors

bengland2:
Do not merge yet, testing, I'll merge as soon as I see it work.

IOPS results in ripsaw-fio-analyzed-result index were way less than sum of per-job (per-process) IOPS results with numjobs=32, this was because the tests in question had per-process IOPS < 1 for 4096-KiB writes, in some cases so int()  truncated them to zero.   In most cases this does not result in such a noticeable error.

### DIFF
--- a/snafu/fio_wrapper/fio_analyzer.py
+++ b/snafu/fio_wrapper/fio_analyzer.py
@@ -82,10 +82,10 @@ class Fio_Analyzer:
                     self.sumdoc[sample][rw][bs_value]["write"] = 0
                     self.sumdoc[sample][rw][bs_value]["read"] = 0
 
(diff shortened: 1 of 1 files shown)


bengland2:
I verified that it worked by comparing results in grafana to results from [analyze-fio-test-results.py](http://nfs-storage01.scalelab.redhat.com/bengland/public/openshift/ripsaw/analyze-fio-test-results.py) for same UUID, it matched exactly, and the right values were also in the ripsaw-fio-analyzed-results index.

bengland2:
please merge now, urgent, I lost the merge button here too.

rsevilla87 (approved):
LGTM!

PR #338, opened 2021-09-08
Put fio local timezone data in UTC+0 timezone to es

inevity:
So that we only change grafana dashboard timezone to the local timezone,
and the timerange selection will always aligin the local timezone


jtaleric (approved):
lgtm

rsevilla87 (approved):
lgtm!

PR #341, opened 2021-09-08
Redis timeout threshold 200 pct

bengland2:
increase redis timeout threshold from 25% of elapsed time to 200% of elapsed time
even with auto-pause feature there is significant variation in elapsed time of pods
also log what redis timeout and redis_timeout_th parameters are.


bengland2:
ready to merge

PR #343, opened 2021-09-09
fs-drift touchup

bengland2:
utilize common code from smallfile to record vfs stats
build without debug RPMs
log inputs


bengland2:
Ryan Drew said CI build system at the time had a problem, hence authorization problem for builds.   Try again when this is fixed, then we should be good to merge.  image build  worked for me with podman.

learnitall:
Hey @bengland2 I merged a fix into master. Can you update your branch with the latest changes from master and push here? The builds will retrigger automatically.

bengland2:
@learnitall doing as you suggested...

bengland2:
@learnitall  it still fails, but perhaps for different reason?  while installing scipy it blew up in numpy, something about requiring python > 3.7, but CI tests exercise python 3.6 thru python3.9.   I don't know of a reason why we need to support python3.6, other than maybe testing on older OCP releases?      What should I do about this?

bengland2:
@learnitall on second thought, it appears only the "arm64" tests failed.   The amd64 ones worked.  Perhaps arm support for numpy and scipy was limited in earlier python versions?   If so, this has nothing to do with this PR so we should merge it, and perhaps skip arm64 tests for earlier python versions?

learnitall:
Thanks for doing that @bengland2 . The arm64 builds are something that we are working towards, need to iron issues with python versions as you mentioned. Something I've got next on my TODO. I believe that your hypothesis is correct.

learnitall (approved):
LGTM

PR #346, opened 2021-09-10
Avoid pushing image in pull_request event

rsevilla87:

### Description

We shouldn't push container images in pull_request events. These kind of events does not have access to GHA secrets. In addition, we should test building images only, pushing them is not a hard requirement for this kind of CI.
Also adding some newlines.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,4 @@
 name: Build Images
-
 on:
   push:
     branches:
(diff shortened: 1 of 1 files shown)


learnitall:
Duplicate of #344.

Additionally the logic in the if statement that you have for `update_manifest` won't work properly. You need to add the `always()` somewhere in there, otherwise the `update_manifest` task will only run if **all** of the images build successfully. In my testing, here is how I found that it needed to be done: https://github.com/cloud-bulldozer/benchmark-wrapper/pull/344/files#diff-d0777657fa3fd81d23aaf7273e58aee453b04e67882517900c56daeef9b3e4c1R165-R169

PR #348, opened 2021-09-10
add directory on container because it cannot be created on runtime

morenod:
It is still failing:

```
2021-09-10T15:12:47Z - ERROR    - MainProcess - trigger_scale: E: Failed to save config file: Failed to write file '/.config/ocm/ocm.json': open /.config/ocm/ocm.json: permission denied
```


rsevilla87 (approved):
LGTM

PR #349, opened 2021-09-12
change uperf_ts to timestamp as other workloads timestamp

ebattat:
### Description
change uperf_ts to timestamp as other workloads timestamp
### Fixes
That current Elasticsearch variable for timestamp is uperf_ts.
The best practice is to use 'timestamp' instead of 'uperf_ts'
Also this is the timestamp variable name in other workloads (stressng, hammerdb)

### DIFF
--- a/snafu/benchmarks/uperf/uperf.py
+++ b/snafu/benchmarks/uperf/uperf.py
@@ -95,7 +95,7 @@ class UperfConfig:
 class UperfStat:
     """Parsed Uperf Statistic."""
 
-    uperf_ts: str
+    timestamp: str
     bytes: int
     norm_byte: int
     ops: int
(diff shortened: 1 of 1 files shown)


snafu/benchmarks/uperf/uperf.py
```diff
@@ -279,7 +279,7 @@ def get_results_from_stdout(stdout: UperfStdout) -> List[UperfStat]:
                 norm_ltcy = ((timestamp - prev_timestamp) / norm_ops) * 1000
 
             datapoint = UperfStat(
-                uperf_ts=datetime.datetime.fromtimestamp(int(timestamp) / 1000).isoformat(),
+                timestamp=datetime.datetime.now(),
```
jtaleric, line 282: This is incorrect. Please review https://github.com/cloud-bulldozer/benchmark-wrapper/blob/79d54b37bfa8fd75283579eebedfba106b23f70c/snafu/benchmarks/uperf/uperf.py#L228-L240

jtaleric (changes requested):
We need to further discuss this, as there are other aspects of the automation that use `uperf_ts` like touchstone that would need to change.

rsevilla87:
@ebattat I'm mostly agree with this change, we should transition from uperf_ts to a more standard field named timestamp. However as @jtaleric this implementation is not right, as the actual timestamp is figured from the uperf log traces.

PR #350, opened 2021-09-13
Determine CPU architecture with $basearch

rsevilla87:

### Description

Thanks to the yum variable $basearch, we can optimize how we use centos8 repositories, and get rid of ppc64le specific repos.

### Fixes

Some arm64 image builds fail in CI because we don't set architecture correctly


learnitall:
I'm not sure if modifying ppc64le images is in our scope, I think they are independent images some other folks are using and maintaining. @jtaleric  can you chime in here on this? If we want to add ppc64le support on our end for our wrappers, I'd like it if we make the main Dockerfiles for our wrappers architecture-independent and just do a `podman build --arch`

learnitall:
Looks like all the dnf steps are running successfully for arm64, this is a great change! Nice! More on the ppc64le- just remember we specifically ignore building ppc64le images in our CI due to this support discrepancy: https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/ci/build_matrix.py#L83

rsevilla87:
> Looks like all the dnf steps are running successfully for arm64, this is a great change! Nice! More on the ppc64le- just remember we specifically ignore building ppc64le images in our CI due to this support discrepancy: https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/ci/build_matrix.py#L83

Yeah, step by step, we can address ppc64le container building in a different PR. Let's try to make _atomic_ changes

learnitall:
Ah I see, yeah that makes sense.  I'm just hesitant about modifying some Dockerfiles that another team is using without their input. But on second thought the change is equivalent pretty much. I'm down.

learnitall (approved):
Got my LGTM

rsevilla87:
> Ah I see, yeah that makes sense. I'm just hesitant about modifying some Dockerfiles that another team is using without their input. But on second thought the change is equivalent pretty much. I'm down.

yeah, let's make the affected people know about changes in ppc64le Dockerfiles once we do them :)

PR #354, opened 2021-09-15
Python 3.9 migration & container build fixes

rsevilla87:

### Description

Most of builds for arm64 amd amd64 archs have been fixed. Some findings of the failed ones:
- We're unable to build oslat container since the packages rt-tests and dumb-init are not longer available in the repos. cc: @smalleni 
- Cyclitest build fails due to broken link to the rt-tests RPM, in addition, we should avoid using static links as builds won't work for multiarch builds. cc @smalleni 
- Hammerdb for arm64 fails due to there's no offial odbc driver for arm64. cc @mkarg75

### DIFF
--- a/snafu/flent_wrapper/Dockerfile
+++ b/snafu/flent_wrapper/Dockerfile
@@ -21,7 +21,7 @@ RUN cd iperf2-code && git checkout bf687b4aac023b303cea08bd1a7248d62ad70f47 && .
 FROM registry.access.redhat.com/ubi8:latest
 
 COPY /snafu/image_resources/centos8-appstream.repo /etc/yum.repos.d/centos8-appstream.repo
-RUN dnf install -y --nodocs git python3-pip iputils redis --enablerepo=centos8-appstream && dnf clean all
+RUN dnf install -y --nodocs git python39-pip iputils redis --enablerepo=centos8-appstream && dnf clean all
 
 RUN pip3 install flent
 
--- a/snafu/flent_wrapper/Dockerfile
+++ b/snafu/flent_wrapper/Dockerfile
@@ -8,12 +8,14 @@ RUN mkdir /tmp/install
 WORKDIR /root
 RUN wget https://github.com/HewlettPackard/netperf/archive/netperf-2.7.0.tar.gz && tar -xzf netperf-2.7.0.tar.gz
 WORKDIR /root/netperf-netperf-2.7.0
+# Workaround: We need to download latest version of config.guess to make configure able to detect other CPU architectures correctly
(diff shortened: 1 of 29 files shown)


jtaleric:
@smalleni @mukrishn ```Status code: 404 for https://www.rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/rt-tests-1.9-2.el8.x86_64.rpm (IP: 195.220.108.108)
error building at STEP "RUN dnf -y install https://www.rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/rt-tests-1.9-2.el8.x86_64.rpm": error while running runtime: exit status 1```

snafu/flent_wrapper/Dockerfile
```diff
@@ -8,8 +8,10 @@ RUN mkdir /tmp/install
 WORKDIR /root
 RUN wget https://github.com/HewlettPackard/netperf/archive/netperf-2.7.0.tar.gz && tar -xzf netperf-2.7.0.tar.gz
 WORKDIR /root/netperf-netperf-2.7.0
+# Workaround: We need to download latest version of config.guess to make configure able to detect other CPU architectures correctly
```
learnitall, line 11: Nice catch! I ran into this as well when debugging some of these builds. Glad to see this was included :)

learnitall (changes requested):
@rsevilla87 since you're at it with upgrading to 3.9, can you also add in an extra line for these Dockerfiles to install dependencies from `requirements/py39-reqs/install.txt`? Also do we have any idea on what to do with the builds that are still failing? It would be great if we could get all these images fixed in this PR. For a temporary workaround on the 404 errors on cyclictest, I figured we could extract the rpm from our current images and store them under `snafu/image_resources`, then do the download from GH.

rsevilla87:
> @rsevilla87 since you're at it with upgrading to 3.9, can you also add in an extra line for these Dockerfiles to install dependencies from `requirements/py39-reqs/install.txt`? Also do we have any idea on what to do with the builds that are still failing? It would be great if we could get all these images fixed in this PR. For a temporary workaround on the 404 errors on cyclictest, I figured we could extract the rpm from our current images and store them under `snafu/image_resources`, then do the download from GH.

I've updated the images for cyclitest and oslat to fedora, as those RPMs are not available in the official repositories anymore. Not sure if we want to move forward from here.

learnitall (approved):
Looks good to me! I'll adjust the build script to add ignores for certain architecture/image combinations (i.e. hammerdb on arm64). We can do the requirement files for the image stuffs.

PR #367, opened 2021-10-29
Update OpenShift Scale to use timeout variable and add additional scale check

dry923:
### Description

Updating the OpenShift scale workload to actually utilize the timeout variable it was able to take as well as add an additional check before completion that we are at the correct scale size.


dry923:
pre-commit.ci autofix

jtaleric (approved):
:shipit:

PR #371, opened 2021-11-02
fix sysbench arm64 build

ewchong:
### Description
`arm64`'s container build was failing

### Fixes
Missing `gcc` and `python38-devel` was causing benchmark-wrapper not to install


jtaleric (approved):
LGTM

Issue #347, opened 2021-09-10
Inconsistent Benchmark and Quay Image Names


learnitall:
Looking at any of the build results after #344 (i.e. https://github.com/cloud-bulldozer/benchmark-wrapper/actions/runs/1221646689), we have some inconsistent names between the wrappers in our code and the wrapper images up on quay. Additionally, some wrappers are missing image repositories altogether. Here are the inconsistencies that I found:

|snafu wrapper module name|snafu wrapper tool name (`--tool`)|quay image name|
|---|---|---|
|iperf|n/a|iperf3|
|ycsb|ycsb|ycsb-server|
|fs_drift|fs-drift|fs-drift|
|dns_perf|dns_perf|No repo on quay|
|upgrade_openshift|upgrade|No repo on quay|

I can add logic into the [build_matrix.py](https://github.com/cloud-bulldozer/benchmark-wrapper/blob/master/ci/build_matrix.py) script in order to take care of this, or we can do some refactoring of our image names. What do we think?

jtaleric:
Yes, if this can be scripted to ensure names stay consistent, lets do it!

dry923:
+1 to consistent naming :-)

learnitall:
Wait, just to clarify because I'm not sure what y'all mean: are we modifying `build_matrix` so that the quay image names stay the same? Or are we modifying the quay image names to be consistent with benchmark-wrapper benchmark names?

learnitall:
Talked internally a bit. Will perform the following changes to maintain consistent naming:

* Rename iperf3 image to iperf
* Rename ycsb-server image to ycsb
* Rename fs-drift image to fs_drift
* Rename fs-drift tool to fs_drift
* Create dns_perf image on quay
* Create upgrade_openshift image on quay
* Rename upgrade tool to upgrade_openshift
* Rename scale tool to scale_openshift

Might take a couple of steps to do this, as these are pretty breaking, but should be simple enough to get done.

chentex:
> Talked internally a bit. Will perform the following changes to maintain consistent naming:
> 
>     
- [ ] Rename iperf3 image to iperf
- [ ] Rename ycsb-server image to ycsb
- [ ] Rename fs-drift image to fs_drift
- [ ] Rename fs-drift tool to fs_drift
- [ ] Create dns_perf image on quay
- [ ] Create upgrade_openshift image on quay
- [ ] Rename upgrade tool to upgrade_openshift
[...]

PR #394, opened 2021-11-17
Updates tag for smallfile image

chentex:
Signed-off-by: Vicente Zepeda Mas <vzepedam@redhat.com>

### Description

This is related to: https://github.com/cloud-bulldozer/benchmark-operator/pull/711

Since the tag is being updated in that PR we need to update it here as well for tests to run

### DIFF
--- a/ci/common.sh
+++ b/ci/common.sh
@@ -25,7 +25,7 @@ function update_operator_image() {
   sed -i "s#${default_ripsaw_image_prefix}/hammerdb:master#${SNAFU_WRAPPER_IMAGE_PREFIX}/hammerdb:${SNAFU_IMAGE_TAG}#g" roles/hammerdb/templates/*
   sed -i "s#${default_ripsaw_image_prefix}/iperf3:latest#${SNAFU_WRAPPER_IMAGE_PREFIX}/iperf3:${SNAFU_IMAGE_TAG}#g" roles/iperf3/templates/*
   sed -i "s#${default_ripsaw_image_prefix}/pgbench:latest#${SNAFU_WRAPPER_IMAGE_PREFIX}/pgbench:${SNAFU_IMAGE_TAG}#g"  roles/pgbench/defaults/main.yml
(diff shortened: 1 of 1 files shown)


sjug (changes requested):
Thanks @chentex,

Addressing some issues that I've seen generally here...
When possible can we make small improvements to the code to reduce smell? 

1. Bash loops - a lot of these `sed` lines follow an identical pattern:
```
WORKLOADS=("hammerdb" "iperf3" "sysbench" "uperf" "vegeta" "scale_openshift" "stressng" "flent" "image_pull" "log_generator")

for workload in "${WORKLOADS[@]}"; do
    sed -i "s#${default_ripsaw_image_prefix}/${workload}:latest#${SNAFU_WRAPPER_IMAGE_PREFIX}/${workload}:${SNAFU_IMAGE_TAG}#g" roles/${workload}/templates/*
done
```
2. One line PRs - far too common in this org, we're changing one tag from `master`. 
Why are `fs-drift` and `hammerdb` still using `master`? Shouldn't they be consistent? If they are/should be consistent shouldn't we pull the tag out as a `${TAG}`?


Out of scope but notable from a quick look at this function:

3. Some image names do not match the role name `fio` vs `fio_distributed`, `ycsb-server` vs `ycsb`. Not sure why this is but they should follow a consistent pattern?

4. Some benchmarks `fs-drift`, `smallfile` have to modify files in templates and roles. We should standardize them (3 of 15 are non-standard).

We need to have some record of tech debt like this that we can work into the backlog at some point. 

/cc @jtaleric @dry923

dry923:
@sjug a couple quick thoughts on this
1/2) re the bash loop I'm 100% on board w/ that assuming we align the images on latest. At this point I don't think there would be any reason to keep than on anything else.
3) There is an existing issue around that https://github.com/cloud-bulldozer/benchmark-wrapper/issues/347

chentex:
Superseded by https://github.com/cloud-bulldozer/benchmark-wrapper/pull/396

PR #401, opened 2021-12-20
handle a case where 1st sample for uperf norm_ops is not 0

mohit-sheth:
### Description
The assumption that the first sample (norm_ops) is always going to be 0 is not true. 
To handle this we exclude first sample from our data

### DIFF
--- a/snafu/benchmarks/uperf/uperf.py
+++ b/snafu/benchmarks/uperf/uperf.py
@@ -273,21 +273,19 @@ class Uperf(Benchmark):
             timestamp, num_bytes, ops = result.timestamp, result.bytes, result.ops
 
             norm_ops = ops - prev_ops
-            if norm_ops == 0:
-                norm_ltcy = 0.0
-            else:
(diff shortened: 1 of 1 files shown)


snafu/benchmarks/uperf/uperf.py
```diff
@@ -273,21 +273,21 @@ def get_results_from_stdout(stdout: UperfStdout) -> List[UperfStat]:
             timestamp, num_bytes, ops = result.timestamp, result.bytes, result.ops
 
             norm_ops = ops - prev_ops
-            if norm_ops == 0:
-                norm_ltcy = 0.0
+            if norm_ops == 0 or prev_timestamp == 0.0:
+                pass
```
mnmehta, line 277: LGTM, perhaps instead of the "pass" use "if norm_ops != 0 and prev_timestamp != 0.0:"

mohit-sheth: makes sense, thanks

PR #404, opened 2022-01-19
Add summary output for uperf

jtaleric:
### Description

Uperf spews raw per-second data to the screen. Instead, we should
provide a summary of the results.

The only issue here, is the user doesn't have the contextg of what the
test was, since it is in XML and read in via a file. This could be later
improved to read in the XML to understand the sort of test.

### DIFF
--- a/snafu/benchmarks/uperf/uperf.py
+++ b/snafu/benchmarks/uperf/uperf.py
@@ -139,41 +139,23 @@ class Uperf(Benchmark):
         # These need help text
         ConfigArgument("--ips", dest="client_ips", env_var="ips", default=""),
         ConfigArgument("--remoteip", dest="remote_ip", env_var="h", default=""),
-        ConfigArgument(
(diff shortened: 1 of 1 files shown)


jtaleric:
Not sure why that check is failing, as it succeeds locally?

```
(rook) jtaleric at rook-mobile in ~/dev/snafu on uperf-summary* $ black snafu/benchmarks/uperf/uperf.py
All done! ✨ 🍰 ✨
1 file left unchanged.

```

mohit-sheth (approved):
lgtm

mohit-sheth:
pre-commit.ci autofix

learnitall:
@jtaleric Snafu's black config is set in the pre-commit config, which isn't picked up by running `black` on its own. To run black with snafu's config, do a `pre-commit run black`.

jtaleric:
Thanks @learnitall!!

PR #410, opened 2022-01-26
updates to coremark-pro and sysbench

ewchong:
### Description
- Sync field names between coremark-pro and sysbench
- Update docs for CoreMark-Pro


learnitall (approved):
LGTM

PR #412, opened 2022-02-24
Assume sysbench is in path

ewchong:
Original implementation had sysbench hardcoded to a path
which breaks on different distros.


learnitall (approved):
LGTM!

PR #415, opened 2022-03-14
Add centos-stream repos, and point centos repos to vault.centos.org.

dagrayvid:
This PR adds 2 repos for centos stream 8, and switches hammerdb to use them. 

It also "fixes" the existing centos8 repos. They currently point to a URL that has been deprecated. Switched to vault.centos.org.

### DIFF
--- a/snafu/hammerdb/Dockerfile
+++ b/snafu/hammerdb/Dockerfile
@@ -1,14 +1,14 @@
 FROM registry.access.redhat.com/ubi8:latest
 
 # install requirements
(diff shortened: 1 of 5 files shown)


rsevilla87:
We've transitioned to Centos stream in benchmark-operator rather than centos vault, I think we should do the same here

dagrayvid:
This can be closed in favor of  #417

dagrayvid:
Closing in favor of  #417

PR #427, opened 2022-05-25
add ocp platform and networktype in scale index

amitsagtani97:
### Description
Add Openshift version and cluster networktype in the scale workload data index


rsevilla87:
There're permission issues, they should be fixed by https://github.com/cloud-bulldozer/benchmark-operator/pull/763

rsevilla87 (approved):
LGTM!

PR #459, opened 2022-10-21
Error out when using archive tool when there are ES connection errors

ewchong:
### Description
There were no errors when running `tool=archive` and there was no connection to ES.

### Fixes
- Log ES errors and exit with a non-zero code when `tool=archive`
- Catch an exception when processing the archive file fails and exit with an error.


jtaleric (approved):
LGTM thanks Ed!

jtaleric:
Hey Ed, please address the lint issues

PR #460, opened 2022-11-02
Clean up remaining pylint issues.

ewchong:
### Description

Got rid of the rest of the pylint errors that were showing up. I split each fix into an individual commit to make it easier to diff.

### Fixes
- Renamed `sysbench` folder to `sysbench_wrapper`
- Added exclusion for `hammerdb` instead of renaming since I don't know how to test it
- Made the `FuncAction` class' `func()` method a static method to get rid of `arguments-differ` from both implementations of it. Tested using `--labels` and was able to error out with the invalid format for the labels.
- Ignore the `arguments-differ` for uperf since it doesn't have a clean up method. 
- Removed two lines in the `.pylintrc` that didn't work.
- Fixed two tests related trivial pylint issues


ewchong:
@jtaleric this cleans up the rest of the pylint issues

PR #461, opened 2022-11-07
Sysbench accept args

ewchong:
### Description
The wrapper for sysbench only took a file for arguments. 

### Fixes
- Added the ability to specify arguments through CLI using `--args=""` and arguments must be surrounded by quotes
- Addressed a couple existing pylint issues.

### DIFF
--- a/snafu/sysbench_wrapper/trigger_sysbench.py
+++ b/snafu/sysbench_wrapper/trigger_sysbench.py
@@ -10,24 +10,31 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+"""
+sysbench benchmark runs
+"""
 
 import logging
 import os
 import re
 import subprocess
+import sys
 from datetime import datetime
 
 logger = logging.getLogger("snafu")
 
 
 class trigger_sysbench:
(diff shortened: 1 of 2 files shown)


snafu/sysbench_wrapper/trigger_sysbench.py
```diff
@@ -40,27 +47,33 @@ def _run_sysbench(self):
...
+        elif self.sysbench_file:
+            config_options = open(self.sysbench_file, encoding="utf8")
+        else:
+            logger.error("Arugments were not specified by file or command line")
+            sys.exit(1)
 
         cmd = "sysbench"
         #  for each option add it to the command line
-        for option in config_file:
+        for option in config_options:
```
jtaleric, line 61: If the user passes a file, all this will be is a pointer to the file, we need to `read` and `split` I would presume?

jtaleric (changes requested):
Did you happen to test with a file? I am not sure that would succeed?

snafu/sysbench_wrapper/trigger_sysbench.py
ewchong: Just tested and it works. This was the code prior to me adding the arguments and that worked.

jtaleric (approved):
ack thanks @ewchong

rsevilla87 (approved):
lgtm!

PR #480, opened 2023-03-22
Allow messages received to be more than sent

morenod:
I have seen this kind of results a few times when running the logging test:

| Messages Sent | Expected Messages | Messages Received
| --- | --- | --- |
| 10800000 | 10800000 | 10800019
| 360000 | 360000 | 360059

This is making the test going to the timeout and setting to false the result.

This can be fixed comparing with >= instead of ==


dry923 (approved):
LGTM

PR #482, opened 2023-03-31
tolerance to allow few missing messages

mukrishn:


snafu/log_generator_wrapper/trigger_log_generator.py
```diff
@@ -285,6 +285,8 @@ def emit_actions(self):
                     messages_received = self._check_kafka(start_time, end_time + self.timeout)
                 if messages_received >= message_count:
                     received_all_messages = True
+                elif self.es_url and messages_received >= message_count-10: # ES misses 1 or 2 messages inconsistently and very negligible, so tolerating 0.00001% loss.
```
dry923, line 288: Do we want to just standardize on allowing 10 missed messages and combine this with the if above?

dry923 (approved):
LGTM!

